PR #4, opened 2020-08-06
PoC Heun's method

mtsokol:
Hi!

I'm a master degree student and I've recently started learning SDE topics, and I came across your project.

For the purpose of learning I've decided to try contributing - in **known issues** it states that Stratonovich methods are missing so I thought about trying that: Here's a draft of Euler counterpart - [Heun's method](https://en.wikipedia.org/wiki/Heun%27s_method). 

If you would have a bit of time for a code review and guidance and it won't be too difficult for me I would like to try contributing. 

It's just a draft with example run in [demo notebook](https://github.com/mtsokol/torchsde/blob/first-stratonovich-solver/examples/demo.ipynb) so without tests yet. What do you think about it?

Do you already have a concrete plan for Stratonovich module? 
Or do you suggest other good-first-issue to start? 

Thank you for any help.


lxuechen:
Hi, 

Thanks for the PR! I think this is a fantastic first step in getting the Stratonovich solvers in place!

Before we move forward, there are two things I want to mention:
1) In order for this code to get merged, you would need to sign a Contributor License Agreement (CLA) per the contributing to google repo policy as hinted by the bot. IIRC, there are two versions: one for individuals and one for corporates. Going through the procedure shouldn't take too long. 

2) Even though this is a simple solver, I'd still recommend writing up some numerical tests to see it's convergence order. I think a reasonable test is that we find a particular SDE, and test the Stratonovich solver's solution with varying step size against the solution from an Ito solver with very small step size that's already implemented in the code base. 

With a proper correction term, a Stratonovich SDE can be converted to Ito form, and hence solved by an Ito solver. We could treat the solution found by the Ito solver as the "ground-truth" and thereafter inspect the convergence order of Heun's method by plotting the errors vs step size on a log-log scale. [Here's](https://github.com/google-research/torchsde/blob/master/diagnostics/srk_diagonal.py) an example I wrote for the SRK solver. 

We don't have to include the code for this particular numerical test, and it might even seem like a slog to do the testing. But I think this is definitely important in the long run, since the last thing any user might hope for is a bug in one of the solvers. 

If you're happy to complete these two steps, I'd be more than excited to move forward with this PR. I will also leave some comments on the code when I start working at some point over the weekend. 

Let me know if there's more that I could help, and thanks again for starting this line of work.

mtsokol:
@googlebot I signed it!

mtsokol:
@lxuechen Thank you for your answer.

That sounds great! 
Let me spend some time on diving into the source code - I'll prepare the test and push it.
Should it be done as mentioned `diagnostic` example or as additional method in `TestSdeint`? Or both?

lxuechen:
I think it'd be nice if we could see a plot on log-log scale just to inspect the rate. We don't necessarily need to include the code for the test, since I'm not sure how much extra stuff that test might need. If the test code isn't too long and roughly follows the format of the other source files in the `diagnostics` folder, I'd probably name it `stratonovich_diagonal.py` and place it in the same folder.

lxuechen:
Here's an example of what a rate-inspection plot would look like. I did this for the diagonal Ito methods, and you can actually run the code in the `diagnostics` folder to obtain this for yourself. The numbers in the parentheses are slopes and correspond to the strong orders of different methods. 

For Heun's method, you should expect to get something that's close to 0.5.  

rate

mtsokol:
@lxuechen Hi! 
I've managed to push first version of the `stratonovich_diagonal.py` diagnostics file. I used problem from `skr_diagonal.py` and run/plot it against euler and analytical solution:




...and the log-log plot:



Shapes of `heun` solutions look good but they seem to be slightly shifted in few cases (Is it OK?).  
Also slope is close to 0.5.

I would be thankful for feedback - I hope that implementation is correct: I looked at formula and existing implementations: 
- [StochasticDiffEq.jl](https://github.com/SciML/StochasticDiffEq.jl/blob/master/src/perform_step/low_order.jl#L63)
- [sdeint](https://github.com/mattja/sdeint/blob/master/sdeint/integrate.py#L229)

This is almost the same, except for `dW` - it above implementations they use the same `dWn` for both stages where here I needed to compute next `I_k` for the second stage to get these solutions. I'm not sure if hadn't got it wrong while reading those implementations. 

I can also polish PR in terms of naming and structure. 

Also I think about minor restructuring - Could `SDEIto` be removed in favor of `BaseSDE` as it only gives `sde_type` (`ForwardSDEIto` and `ForwardSDEStratonivich` will inherit from `BaseSDE` and provide `sde_type` there)? 
Then all problems from `problems.py` can inherit from `BaseSDE` and be solved with both methods by providing concrete `sde_type` when instantiated. WDYT?

lxuechen:
Hi,

Thanks for the updates! I just checked the code and the solver seems good. 

Now regarding the rate inspection, there's a technical issue which is for an Ito SDE and Stratonovich SDE, their analytical solutions would be different, even when their drifts and diffusions are the same. So I am guessing that's the reason to the rather uniform vertical difference between Euler and Heun as you vary the step size. 

One way to properly do the error analysis is to use a conversion rule to convert the Strat SDE to Ito form and then use the Euler solver based on Ito's form. 

Now that I think about it, it might require more work than I originally thought, and things can become messy. I'm currently working on setting up the Stratonovich utils and have plans of doing the rate inspection properly. On the other hand, I don't want to leave your contributions unnoticed. So here's a plan that I propose:

It would be nice if you could only include the following files
- `torchsde/__init__.py`
- `torchsde/_core/base_sde.py`
- `torchsde/_core/methods_strat/__init__.py `
- `torchsde/_core/methods_strat/diagonal/__init__.py`
- `torchsde/_core/methods_strat/diagonal/heun.py `
- `torchsde/_core/sdeint.py ` 
- `torchsde/_core/settings.py` 

I can take it up from there and set up all the stuff that's relevant for the analysis. 

What do you think about this plan?

mtsokol:
@lxuechen Thank you for your feedback!

That sounds good to me - I deleted `diagnostic` file leaving only 7 files requested in PR.

So what can I work on next? 
Maybe I can work on next method - Milstein method for Stratonovich problems? (I will finish `gdg_prod` in `ForwardSDEStratonovich` for that purpose).
If not this maybe something different?

torchsde/_core/methods_strat/diagonal/heun.py
```diff
@@ -0,0 +1,60 @@
...
+        f_eval = self.sde.f(t0, y0)
+        g_eval = self.sde.g(t0, y0)
+        g_prod_eval = self.sde.g_prod(g_eval, I_k)
+        y1 = [
+            y0_i + f_eval_i * dt + g_prod_eval_i
+            for y0_i, f_eval_i, g_prod_eval_i in zip(y0, f_eval, g_prod_eval)
+        ]
+        
+        t1 = t0 + dt
+        I_k = [(bm_next - bm_cur).to(y1[0]) for bm_next, bm_cur in zip(self.bm(t1 + dt), self.bm(t1))]
```
patrick-kidger, line 38: This looks wrong to me. You should only expect to evaluate the Brownian motion once.

torchsde/_core/methods_strat/diagonal/heun.py
```diff
@@ -0,0 +1,60 @@
...
+        y2 = [
+            y0_i + 0.5 * f_eval_i * dt + g_prod_eval_i
+            for y0_i, f_eval_i, g_prod_eval_i in zip(y0, f_eval_next_step_sum, g_prod_eval_next_step)
+        ]
+
+        return t1, y2
+
+    @property
+    def strong_order(self):
+        return 0.5
```
patrick-kidger, line 56: This should be 1.0.

I notice that you get 0.5 in your tests; this is probably because of the earlier bug with the Brownian motion. (Or perhaps because you're comparing Ito against Stratonovich?)

It doesn't help btw that there are actually two different methods called "Heun". One with strong order 0.5, and one with strong order 1.0. The one that it looks like you're aiming to implement has strong order 1.0.

torchsde/_core/methods_strat/diagonal/heun.py
patrick-kidger: Search for "Heun" in https://royalsocietypublishing.org/doi/pdf/10.1098/rspa.2003.1247
which might be helpful.

mtsokol: Thanks for these materials! Ok, now I see what I got wrong - I needlessly added another `bm` evaluation (which got me better solution without proper conversion) - removed it. So as mentioned the error was due to the lack of conversion between Ito and Strat - For problem `Ex2` I calculated and added correction for `f`:
```python
def f_corr(self, t, y):
    return 0.5 * (-self.p ** 4. * torch.sin(y) * torch.cos(y) ** 5. * (torch.cos(y) ** 2. - 3. * torch.sin(y) ** 2.))
```
And now it solves it way better:



And I got such rate:

mtsokol: And as you mention I will resubmit it to `dev` branch to be up to date with current API design changes

patrick-kidger: Could you share your derivation of the correction? Converting from Ito to Stratonovich, the modified drift is `O(f, g.dg) = O(p^2)` but what you've written is `O(p^4)`.

Also, the rate in Heun (the slope) in your graph is now about 0.1, which is very wrong.

Your implementation now looks correct, though, so I suspect that may be because of the drift being wrong.

torchsde/_core/methods_strat/diagonal/heun.py
```diff
@@ -0,0 +1,59 @@
...
+        ]
+        
+        t1 = t0 + dt
+
+        f_eval_next_step_sum = [
+            a + b for a, b in zip(f_eval, self.sde.f(t1, y1))
+        ]
+        g_eval_next_step_sum = [
+            (a + b) / 2.0 for a,b in zip(g_eval, self.sde.g(t1, y1))
+        ]
```
patrick-kidger, line 44: Nitpick: can you put these sums inside the `y2` calculation, just so that we don't have to materialise extra lists?

torchsde/_core/methods_strat/diagonal/heun.py
```diff
@@ -0,0 +1,59 @@
...
+            for y0_i, f_eval_i, g_prod_eval_i in zip(y0, f_eval, g_prod_eval)
+        ]
+        
+        t1 = t0 + dt
+
+        f_eval_next_step_sum = [
+            a + b for a, b in zip(f_eval, self.sde.f(t1, y1))
+        ]
+        g_eval_next_step_sum = [
+            (a + b) / 2.0 for a,b in zip(g_eval, self.sde.g(t1, y1))
```
patrick-kidger, line 43: Nitpick: can you switch `/ 2` to `* 0.5`? Off the top of my head I don't know if Python/PyTorch is smart enough to optimise static divisors, but any unnecessary divisions will slow things down.

torchsde/_core/methods_strat/diagonal/heun.py
```diff
@@ -0,0 +1,59 @@
...
+        ]
+        g_prod_eval_next_step = self.sde.g_prod(g_eval_next_step_sum, I_k)
+        y2 = [
```
patrick-kidger, line 46: Nitpick: can this be called `y1`, and the intermediate point earlier called `y0_prime` or similar? Otherwise returning `t1, y2` looks wrong.

torchsde/_core/methods_strat/diagonal/heun.py
mtsokol: OK again my error 😅, sorry! Did Strat => Ito instead of the opposite. 

So current rate is (Is it correct?):

And all plots look like this:

And my correction for `Ex2`:

patrick-kidger: Excellent! The rate now looks correct and the Ito/Strat correction matches my derivation.

torchsde/_core/methods_strat/diagonal/heun.py
mtsokol: Done in newer PR 👍

torchsde/_core/methods_strat/diagonal/heun.py
mtsokol: Done in newer PR 👍

torchsde/_core/methods_strat/diagonal/heun.py
mtsokol: Done in newer PR 👍

patrick-kidger:
Closing in favour of #24

Issue #8, opened 2020-08-09
Operate on tensors rather than tuple-of-tensors?


patrick-kidger:
Hi Xuechen - I appreciate I'm bombarding this repository with a lot of issues/PRs.

Something we've recently found over in [torchdiffeq#115](https://github.com/rtqichen/torchdiffeq/pull/115) is that it's noticeably more efficient to have the internals of the library operate on tensors, rather than tuples-of-tensors, and have the adjoint method simply `torch.cat` its different channels together.

Of course this also makes the code a whole lot cleaner, and I expect easier to develop going forward.

Not something I've got time to put together a PR for I'm afraid, but I thought I'd flag this up as something you might be interested in over here.

lxuechen:
Hi Patrick,

Thanks for the note! It's interesting that using `torch.cat` for the output can speed things up by much. I definitely learned something here. 

I'll try to patch this up over a weekend in the future, and feel free to open up other issues even if you don't have cycles to work on them. Really appreciate it!

patrick-kidger:
No problem. More issues to follow!

patrick-kidger:
Done in #37.

PR #19, opened 2020-08-16
Dev kidger2

patrick-kidger:
Quite a lot of bugfixes, mostly.

<details>
<summary>List</summary>

- Updated diagnostics.
- Removed trapezoidal_approx
- Fixed error messages for wrong methods etc.
- Can now call BrownianInterval with a single value.
- Fixed bug in BrownianInterval that it was always returning 0!
- There's now a 2-part way of getting levy area: it has to be set as available during `__init__`, and then specified that it's wanted during `__call__`. This allows one to create general Brownian motions that can be used in multiple solvers, and have each solver call just the bits it wants.
- Bugfix spacetime -> space-time
- Improved checking in check_contract
- Various minor tidy-ups
- Can use Brownian* with any Levy area sufficient for the solver, rather than just the minimum the solver needs.
- Fixed using bm=None in sdeint and sdeint_adjoint, so that it creates an appropriate BrownianInterval. This also makes method='srk' easy.
</details>

The diagnostics and `sdeint(bm=None)` both use BrownianInterval just because that's the only one offering `U` at the moment.

Also, whilst its not quite a formal test, the diagnostics seem positive about BrownianInterval; in particular it looks like SRK vs analytic is consistent, which implies that the space-time Levy area is likely correct wrt increment.

Btw, what name would you use for `U`? I'd describe it as "a cross term in the second order signature", but that's not very snappy and I'd like to replace `LEVY_AREA_APPROXIMATIONS.spacetime` with a more appropriate name.

### DIFF
--- a/examples/latent_sde.py
+++ b/examples/latent_sde.py
@@ -27,7 +27,7 @@ from torch import nn, optim
 from torch.distributions import Normal, Laplace, kl_divergence
 
 from examples import utils
-from torchsde import sdeint, sdeint_adjoint, SDEIto, BrownianPath
+from torchsde import sdeint, sdeint_adjoint, SDEIto, BrownianInterval
 
 Data = namedtuple('Data', ['ts_', 'ts_ext_', 'ts_vis_', 'ts', 'ts_ext', 'ts_vis', 'ys', 'ys_'])
 
@@ -83,15 +83,11 @@ class LatentSDE(SDEIto):
         logqp0 = kl_divergence(qy0, py0).sum(1).mean(0)  # KL(time=0).
 
         if args.adjoint:
-            zs, logqp = sdeint_adjoint(
-                self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive, rtol=args.rtol,
-                atol=args.atol
-            )
+            zs, logqp = sdeint_adjoint(self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive,
+                                       rtol=args.rtol, atol=args.atol)
         else:
-            zs, logqp = sdeint(
-                self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive, rtol=args.rtol,
-                atol=args.atol
-            )
+            zs, logqp = sdeint(self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive,
+                               rtol=args.rtol, atol=args.atol)
         logqp = logqp.sum(0).mean(0)
         log_ratio = logqp0 + logqp  # KL(time=0) + KL(path).
 
@@ -175,7 +171,8 @@ def main():
 
     # Fix seed for the random draws used in the plots.
     eps = torch.randn(vis_batch_size, 1).to(device)
-    bm = BrownianPath(t0=ts_vis[0], w0=torch.zeros(vis_batch_size, 1).to(device))
+    bm = BrownianInterval(t0=ts_vis[0], t1=ts_vis[-1], shape=(vis_batch_size, 1), dtype=torch.float32, device=device,
+                          levy_area_approximation='space-time')  # We need space-time Levy area to use the SRK solver
 
     # Model.
     model = LatentSDE().to(device)
--- a/examples/latent_sde.py
+++ b/examples/latent_sde.py
@@ -82,16 +82,15 @@ class LatentSDE(SDEIto):
         py0 = Normal(loc=self.py0_mean, scale=self.py0_std)
         logqp0 = kl_divergence(qy0, py0).sum(1).mean(0)  # KL(time=0).
 
-        # `trapezoidal_approx` is for SRK. Disabling it gives better performance.
         if args.adjoint:
             zs, logqp = sdeint_adjoint(
                 self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive, rtol=args.rtol,
-                atol=args.atol, options={'trapezoidal_approx': False}
+                atol=args.atol
             )
         else:
             zs, logqp = sdeint(
                 self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive, rtol=args.rtol,
-                atol=args.atol, options={'trapezoidal_approx': False}
+                atol=args.atol
             )
         logqp = logqp.sum(0).mean(0)
         log_ratio = logqp0 + logqp  # KL(time=0) + KL(path).
--- a/torchsde/_brownian/brownian_interval.py
+++ b/torchsde/_brownian/brownian_interval.py
@@ -446,7 +446,7 @@ class BrownianInterval(_Interval, base_brownian.BaseBrownian):
         self.levy_area_approximation = levy_area_approximation
 
         # Precompute these as we don't want to spend lots of time checking strings in hot loops.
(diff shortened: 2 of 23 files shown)


torchsde/_brownian/brownian_interval.py
```diff
@@ -467,12 +468,15 @@ def increment_and_space_time_levy_area_cache(interval):
...
+    # TODO: pick better names for return_U, return_A. A is clearly 'levy_area', but U?
+    def __call__(self, ta, tb=None, return_U=False, return_A=False):
+        if tb is None:
+            ta, tb = self.start, ta
         ta = float(ta)
         tb = float(tb)
         # Can get queries just inside and outside the specified region in SDE solvers; we just clamp.
-        ta = min(self.start, max(ta, self.end))
-        tb = min(self.start, max(tb, self.end))
+        ta = max(self.start, min(ta, self.end))
```
lxuechen, line 478: This seems like it will fail silently when `ta` or `tb` is out of range. Typically, it shouldn't be a big problem, since we don't expect values that are out of range by too much.

But could we at least raise a warning or something like that, just to defend against the imaginary adversarial user?

torchsde/_core/methods/srk.py
```diff
@@ -32,11 +32,14 @@
 
 
 class SRK(base_solver.BaseSDESolver):
+    # TODO: should the strong order be 2.0 for additive noise? Numerically it looks like it.
     strong_order = 1.5
     weak_order = 1.5
     sde_type = SDE_TYPES.ito
     noise_types = (NOISE_TYPES.additive, NOISE_TYPES.diagonal, NOISE_TYPES.scalar)
-    levy_area_approximation = LEVY_AREA_APPROXIMATIONS.spacetime
+    levy_area_approximations = [LEVY_AREA_APPROXIMATIONS.spacetime,
```
lxuechen, line 40: Could we rename `spacetime` to `space_time`? It matches better with the underlying str.

examples/latent_sde.py
```diff
@@ -82,16 +82,15 @@ def forward(self, ts, batch_size, eps=None):
         py0 = Normal(loc=self.py0_mean, scale=self.py0_std)
         logqp0 = kl_divergence(qy0, py0).sum(1).mean(0)  # KL(time=0).
 
-        # `trapezoidal_approx` is for SRK. Disabling it gives better performance.
         if args.adjoint:
             zs, logqp = sdeint_adjoint(
                 self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive, rtol=args.rtol,
-                atol=args.atol, options={'trapezoidal_approx': False}
+                atol=args.atol
```
lxuechen, line 88: Since this super long input arg is removed, could we put atol to the line above to reduce code length?

torchsde/_brownian/modified.py
```diff
@@ -45,15 +43,39 @@ def levy_area_approximation(self):
         return self.base_brownian.levy_area_approximation
 
 
+# TODO: these checks are very inelegant. Is there a better interface to Brownian motion?
```
lxuechen, line 46: I think this is alright for now. We can come back at a later time.

lxuechen:
> Btw, what name would you use for U? I'd describe it as "a cross term in the second order signature", but that's not very snappy and I'd like to replace LEVY_AREA_APPROXIMATIONS.spacetime with a more appropriate name.

I'm somewhat unsure about this. I think we could leave it as `space-time` for now unless you have a better idea. I was gonna call it "integrated_brownian", but this term is too long and not very accurate.

lxuechen:
As a side comment, did you make sure all of the code in `diagnostics`, `examples` still run? 

I can fix the tests later on.

lxuechen:
I realized this PR might conflict with the PR in #20 slightly. I can merge this after the comments are resolved, and feel free to leave plenty of comments on my PR. 

For now, I'm moving forward with the adjoint for general noise Strat SDE.

torchsde/_brownian/brownian_interval.py
patrick-kidger: That would raise a warning as a result of normal solver behaviour as well, though.

FWIW I'm inclined to actually regard this as bug in the solvers. We don't have to step past endpoint and interpolate back: we can just truncate our final step to exactly match the endpoint.

Then this line can be switched for an error.

torchsde/_core/methods/srk.py
patrick-kidger: Agreed

examples/latent_sde.py
patrick-kidger: Agreed

torchsde/_brownian/modified.py
patrick-kidger: Yeah, there's definitely more important things. Things should also simplify a lot once the tupled logic gets replaced by cat-tensor logic as per #8.

patrick-kidger:
> 
> 
> > Btw, what name would you use for U? I'd describe it as "a cross term in the second order signature", but that's not very snappy and I'd like to replace LEVY_AREA_APPROXIMATIONS.spacetime with a more appropriate name.
> 
> I'm somewhat unsure about this. I think we could leave it as `space-time` for now unless you have a better idea. I was gonna call it "integrated_brownian", but this term is too long and not very accurate.

Area under curve - `auc` - maybe?


> 
> 
> As a side comment, did you make sure all of the code in `diagnostics`, `examples` still run?
> 
> I can fix the tests later on.

I checked that `benchmark/brownian.py`, `diagnostics/ito_additive.py` and `diagnostics/ito_diagonal.py` run. I found that `diagnostics/ito_scalar.py` takes a long time and eats up memory, even when using BrownianPath rather than BrownianInterval. I've not dug into why that is. I think some of the examples might need updating, but I'll check those and report back.

patrick-kidger:
`latent_sde.py` seems to run correctly. Other than evaluting BrownianInterval outside of t0, t1, I think that's everything.

torchsde/_brownian/modified.py
lxuechen: Agree

torchsde/_brownian/brownian_interval.py
lxuechen: > FWIW I'm inclined to actually regard this as bug in the solvers. We don't have to step past endpoint and interpolate back: we can just truncate our final step to exactly match the endpoint. Then this line can be switched for an error.

I agree with this. 

Overall, I think the expected behavior would be something like this
```python
bm = BrownianInterval(t0=0., t1=1.0, ...)
y0 = torch.zeros(3, 4)
ts = torch.tensor([0., 2.])
sdeint(y0, ts, bm=bm)  # Raises warning/error eventually
```

What do you think? Would you like to give it a go?

lxuechen:
Ok, the merge looks good. 

Issues we should still keep track of are
- naming of U
- warning/error message for bm when query is outside range

torchsde/_brownian/brownian_interval.py
patrick-kidger: Can do. I'll do it as a separate PR though, so that you can work off of this one in your branches.

patrick-kidger: (Hah, you ninja'd my comment with a merge anyway.)

lxuechen: Sorry, was trying to move fast. Thought you'd do it separately. Anyway, I'll be more careful next time!

Issue #26, opened 2020-08-17
Make `dtype` of BrownianInterval constructor an optional argument


lxuechen:
Could we make this argument an optional one so that users are freed from typing this in when they explicitly initialize BrownianInterval? An optional could be `torch.get_default_dtype()`.

patrick-kidger:
Resolved in #28

Issue #25, opened 2020-08-17
Bug in BrownianInterval


lxuechen:
This bug relates to my concern that there's virtually no test for BrownianInterval at the moment, and in retrospect it shouldn't have been merged into dev without testing out the aspects I mentioned in #15. Shape tests are basic, though it is helpful at most times. The change of API for the solvers also breaks backward compatibility.
 
The data structure couldn't produce `A` with the correct shape

```python
import torch

from torchsde import BrownianInterval

device = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')

batch_size, d, m = 16, 3, 2
t0, t1 = 0.0, 0.3
dtype = torch.get_default_dtype()

bm_general = BrownianInterval(
    t0=t0, t1=t1, shape=(batch_size, m), dtype=dtype, device=device, levy_area_approximation='foster'
)
W, U, A = bm_general(0.0, 0.1)  # Fails.
print(W.size())
print(U.size())
print(A.size())
```

patrick-kidger:
Resolved in #28

Issue #10, opened 2020-08-09
Calling `BrownianTree` with `t > t1` returns `None`


patrick-kidger:
Repro:
```
torchsde.BrownianTree(t0=torch.tensor(0.), w0=torch.tensor(0.))(2.)
```
Not helped by the implicit assumption that `t1 = t0 + 1`.

lxuechen:
This seems to be an issue caused by `t` > `t1` + `safety`, where `safety` is one of the arguments for initialization. I agree that the user shouldn't be obliged to always set the safety factor precisely. 

Will send in a fix shortly.

patrick-kidger:
CLosing as resolvedi n #28, as querying with t > t1 is no longer possible.

Issue #9, opened 2020-08-09
Global entropy doesn't determine sampled values outside of [t0, t1] for `BrownianTree`


lxuechen:
Following the [note](https://github.com/google-research/torchsde/pull/7#issuecomment-671078247), the `entropy` argument passed to initialize `BrownianTree` doesn't govern the random numbers sampled outside the predetermined range [t0, t1]. 

Queries of this type is rate. Though, at the very least, there should be some warning or documentation on this.

patrick-kidger:
Closing as resolved in #28, as querying outside [t0, t1] is no longer possible.

PR #28, opened 2020-08-17
BrownianInterval tests + bugfixes

patrick-kidger:
Many things on `BrownianInterval`:
- Added tests!
- Fixed various bugs (#25)
- Sped things up slightly
- dtype/device now optional arguments (#26)
- Fixed getting RecursionErrors with very large cache sizes

Fixed behaviour for calling `BrownianInterval` and `BrownianTree` for `t` outside of their range (it now clamps and raises a warning) (#9, #10)

Fixed bug that prevented `ForwardSDE.g_prod` from being called. Tided it slightly.

Fixed SDE solvers stepping outside the region of integration.

Btw @lxuechen I don't actually follow what you were saying about there being a bug in `ReverseBrownian`?

### DIFF
--- a/tests/test_brownian_interval.py
+++ b/tests/test_brownian_interval.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Test the two `BrownianTree`.
+"""Test the `BrownianInterval``.
 
 The suite tests both running on CPU and CUDA (if available).
 """
--- a/tests/test_brownian_interval.py
+++ b/tests/test_brownian_interval.py
@@ -33,7 +33,7 @@ torch.set_default_dtype(torch.float64)
 
 D = 3
 SMALL_BATCH_SIZE = 16
-LARGE_BATCH_SIZE = 16384
+LARGE_BATCH_SIZE = 32000
 REPS = 3
 LARGE_REPS = 500
 ALPHA = 0.001
@@ -42,82 +42,129 @@ ALPHA = 0.001
 devices = [cpu, gpu] = [torch.device('cpu'), torch.device('cuda')]
 
 
-def _setup(device, batch_size):
+def _setup(device, levy_area_approximation, shape):
     t0, t1 = torch.tensor([0., 1.], device=device)
     ta = torch.rand([], device=device)
     tb = torch.rand([], device=device)
     ta, tb = min(ta, tb), max(ta, tb)
-    bm = torchsde.BrownianInterval(t0=t0, t1=t1, shape=(batch_size, D), dtype=torch.float64, device=device)
+    bm = torchsde.BrownianInterval(t0=t0, t1=t1, shape=shape, device=device,
+                                   levy_area_approximation=levy_area_approximation)
     return ta, tb, bm
 
 
-@pytest.mark.parametrize("device", devices)
-def test_shape(device):
-    if device == gpu and not torch.cuda.is_available():
-        pytest.skip(msg="CUDA not available.")
-
-    ta, tb, bm = _setup(device, SMALL_BATCH_SIZE)
-    sample1 = bm(ta)
-    sample2 = bm(tb)
-    sample3 = bm(ta, tb)
-    assert sample1.shape == sample2.shape == sample3.shape == (SMALL_BATCH_SIZE, D)
-
-    ta, tb, bm = _setup(device, SMALL_BATCH_SIZE)
-    # Query interval before increment
-    sample3 = bm(ta, tb)
-    sample1 = bm(ta)
-    sample2 = bm(tb)
-    assert sample1.shape == sample2.shape == sample3.shape == (SMALL_BATCH_SIZE, D)
+def _levy_returns():
+    yield "none", False, False
+    yield "space-time", False, False
+    yield "space-time", True, False
+    for levy_area_approximation in ('davie', 'foster'):
+        for return_U in (True, False):
+            for return_A in (True, False):
+                yield levy_area_approximation, return_U, return_A
 
 
 @pytest.mark.parametrize("device", devices)
-def test_determinism_simple(device):
+@pytest.mark.parametrize("levy_area_approximation, return_U, return_A", _levy_returns())
+def test_shape(device, levy_area_approximation, return_U, return_A):
     if device == gpu and not torch.cuda.is_available():
         pytest.skip(msg="CUDA not available.")
 
-    ta, tb, bm = _setup(device, SMALL_BATCH_SIZE)
-    vals = [bm(ta, tb) for _ in range(REPS)]
+    for shape, A_shape in (((SMALL_BATCH_SIZE, D), (SMALL_BATCH_SIZE, D, D)),
+                           ((SMALL_BATCH_SIZE,), (SMALL_BATCH_SIZE,)),
+                           ((), ())):
+        ta, tb, bm = _setup(device, levy_area_approximation, shape)
+        sample1 = bm(ta, return_U=return_U, return_A=return_A)
+        sample2 = bm(tb, return_U=return_U, return_A=return_A)
+        sample3 = bm(ta, tb, return_U=return_U, return_A=return_A)
+        shapes = []
+        A_shapes = []
+        for sample in (sample1, sample2, sample3):
+            if return_U:
+                if return_A:
+                    W1, U1, A1 = sample
+                    shapes.append(W1.shape)
+                    shapes.append(U1.shape)
+                    A_shapes.append(A1.shape)
+                else:
+                    W1, U1 = sample
+                    shapes.append(W1.shape)
+                    shapes.append(U1.shape)
+            else:
+                if return_A:
+                    W1, A1 = sample
+                    shapes.append(W1.shape)
+                    A_shapes.append(A1.shape)
+                else:
+                    W1 = sample
+                    shapes.append(W1.shape)
+
+        for shape_ in shapes:
+            assert shape_ == shape
+        for shape_ in A_shapes:
+            assert shape_ == A_shape
+
+
+@pytest.mark.parametrize("device", devices)
+@pytest.mark.parametrize("levy_area_approximation, return_U, return_A", _levy_returns())
+def test_determinism_simple(device, levy_area_approximation, return_U, return_A):
+    if device == gpu and not torch.cuda.is_available():
+        pytest.skip(msg="CUDA not available.")
+
+    ta, tb, bm = _setup(device, levy_area_approximation, (SMALL_BATCH_SIZE, D))
+    vals = [bm(ta, tb, return_U=return_U, return_A=return_A) for _ in range(REPS)]
     for val in vals[1:]:
-        assert torch.allclose(val, vals[0])
+        if torch.is_tensor(val):
+            val = (val,)
+        if torch.is_tensor(vals[0]):
+            val0 = (vals[0],)
+        else:
+            val0 = vals[0]
+        for v, v0 in zip(val, val0):
+            assert (v == v0).all()
 
 
 @pytest.mark.parametrize("device", devices)
-def test_determinism_large(device):
+@pytest.mark.parametrize("levy_area_approximation, return_U, return_A", _levy_returns())
+def test_determinism_large(device, levy_area_approximation, return_U, return_A):
     if device == gpu and not torch.cuda.is_available():
         pytest.skip(msg="CUDA not available.")
 
-    ta, tb, bm = _setup(device, SMALL_BATCH_SIZE)
+    ta, tb, bm = _setup(device, levy_area_approximation, (SMALL_BATCH_SIZE, D))
     cache = {}
     for _ in range(LARGE_REPS):
         ta_ = torch.rand_like(ta)
         tb_ = torch.rand_like(tb)
         ta_, tb_ = min(ta_, tb_), max(ta_, tb_)
-        val = bm(ta_, tb_)
-        cache[ta_, tb_] = val.detach().clone()
+        val = bm(ta_, tb_, return_U=return_U, return_A=return_A)
+        if torch.is_tensor(val):
+            val = (val,)
+        cache[ta_, tb_] = tuple(v.detach().clone() for v in val)
 
     cache2 = {}
     for ta_, tb_ in cache:
-        val = bm(ta_, tb_)
-        cache2[ta_, tb_] = val.detach().clone()
+        val = bm(ta_, tb_, return_U=return_U, return_A=return_A)
+        if torch.is_tensor(val):
+            val = (val,)
+        cache2[ta_, tb_] = tuple(v.detach().clone() for v in val)
 
     for ta_, tb_ in cache:
-        assert (cache[ta_, tb_] == cache2[ta_, tb_]).all()
+        for v1, v2 in zip(cache[ta_, tb_], cache2[ta_, tb_]):
+            assert (v1 == v2).all()
 
 
 @pytest.mark.parametrize("device", devices)
-def test_normality_simple(device):
+@pytest.mark.parametrize("levy_area_approximation", ['none', 'space-time', 'davie', 'foster'])
+def test_normality_simple(device, levy_area_approximation):
     if device == gpu and not torch.cuda.is_available():
         pytest.skip(msg="CUDA not available.")
 
     t0_, t1_ = 0.0, 1.0
     t0, t1 = torch.tensor([t0_, t1_], device=device)
-    eps = 1e-5
     for _ in range(REPS):
         W = torch.tensor(npr.randn(), device=device).repeat(LARGE_BATCH_SIZE)
-        bm = torchsde.BrownianInterval(t0=t0, t1=t1, W=W)
+        bm = torchsde.BrownianInterval(t0=t0, t1=t1, W=W, levy_area_approximation=levy_area_approximation)
 
         for _ in range(REPS):
-            t_ = npr.uniform(low=t0_ + eps, high=t1_ - eps)
+            t_ = npr.uniform(low=t0_, high=t1_)
             samples = bm(t_)
             samples_ = samples.cpu().detach().numpy()
 
@@ -130,13 +177,44 @@ def test_normality_simple(device):
 
 
 @pytest.mark.parametrize("device", devices)
+@pytest.mark.parametrize("levy_area_approximation", ['none', 'space-time', 'davie', 'foster'])
+def test_normality_conditional(device, levy_area_approximation):
+    if device == gpu and not torch.cuda.is_available():
+        pytest.skip(msg="CUDA not available.")
+
+    t0_, t1_ = 0.0, 1.0
+    t0, t1 = torch.tensor([t0_, t1_], device=device)
+    for _ in range(REPS):
+        bm = torchsde.BrownianInterval(t0=t0, t1=t1, shape=(LARGE_BATCH_SIZE,),
+                                       levy_area_approximation=levy_area_approximation)
+
+        for _ in range(REPS):
+            t_ = npr.uniform(low=t0_, high=t1_)
+            ta = npr.uniform(low=t0_, high=t1_)
+            tb = npr.uniform(low=t0_, high=t1_)
+            ta, t_, tb = sorted([t_, ta, tb])
+
+            increment = bm(ta, tb).cpu().detach().numpy()
+            sample = bm(ta, t_).cpu().detach().numpy()
+
+            mean = increment * (t_ - ta) / (tb - ta)
+            std = np.sqrt((tb - t_) * (t_ - ta) / (tb - ta))
+            rescaled_sample = (sample - mean) / std
+
+            _, pval = kstest(rescaled_sample, 'norm')
+            assert pval >= ALPHA
+
+
+@pytest.mark.parametrize("device", devices)
+@pytest.mark.parametrize("levy_area_approximation", ['none', 'space-time', 'davie', 'foster'])
 @pytest.mark.parametrize("random_order", [False, True])
-def test_continuity(device, random_order):
+def test_continuity(device, levy_area_approximation, random_order):
     if device == gpu and not torch.cuda.is_available():
         pytest.skip(msg="CUDA not available.")
 
     ts = torch.linspace(0., 1., 10000, device=device)
-    bm = torchsde.BrownianInterval(t0=ts[0], t1=ts[-1], shape=(), dtype=ts.dtype, device=device)
+    bm = torchsde.BrownianInterval(t0=ts[0], t1=ts[-1], shape=(), device=device,
+                                   levy_area_approximation=levy_area_approximation)
     vals = torch.empty_like(ts)
     i_ = torch.arange(len(ts), device=device)
     if random_order:
@@ -150,12 +228,14 @@ def test_continuity(device, random_order):
         last_val = val
 
 
+@pytest.mark.skip("BrownianInterval.to not supported")
 @pytest.mark.parametrize("device", devices)
-def test_to_dtype(device):
+@pytest.mark.parametrize("levy_area_approximation", ['none', 'space-time', 'davie', 'foster'])
+def test_to_dtype(device, levy_area_approximation):
     if device == gpu and not torch.cuda.is_available():
         pytest.skip(msg="CUDA not available.")
 
-    ta, tb, bm = _setup(device, SMALL_BATCH_SIZE)
+    ta, tb, bm = _setup(device, levy_area_approximation, (SMALL_BATCH_SIZE, D))
     tc = torch.rand_like(ta)
     td = torch.rand_like(ta)
     tc, td = min(tc, td), max(tc, td)
@@ -187,11 +267,13 @@ def test_to_dtype(device):
     assert torch.allclose(w4, w4_.double())
 
 
-def test_to_device():
+@pytest.mark.skip("BrownianInterval.to not supported")
+@pytest.mark.parametrize("levy_area_approximation", ['none', 'space-time', 'davie', 'foster'])
+def test_to_device(levy_area_approximation):
     if not torch.cuda.is_available():
         pytest.skip(msg="CUDA not available.")
 
-    ta, tb, bm = _setup(cpu, SMALL_BATCH_SIZE)
+    ta, tb, bm = _setup(cpu, levy_area_approximation, (SMALL_BATCH_SIZE, D))
     tc = torch.rand_like(ta)
     td = torch.rand_like(ta)
     tc, td = min(tc, td), max(tc, td)
--- a/tests/test_brownian_interval.py
+++ b/tests/test_brownian_interval.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Test the `BrownianInterval``.
+"""Test `BrownianInterval`.
 
 The suite tests both running on CPU and CUDA (if available).
 """
--- a/tests/test_brownian_interval.py
+++ b/tests/test_brownian_interval.py
@@ -20,7 +20,7 @@ import sys
 
 sys.path = sys.path[1:]  # A hack so that we always import the installed library.
 
-import numpy as np
+import math
 import numpy.random as npr
 import torch
 from scipy.stats import norm, kstest
@@ -33,11 +33,11 @@ torch.set_default_dtype(torch.float64)
 
 D = 3
 SMALL_BATCH_SIZE = 16
-LARGE_BATCH_SIZE = 32000
+LARGE_BATCH_SIZE = 131072
 REPS = 3
 LARGE_REPS = 500
 ALPHA = 0.001
-
+POOL_SIZE = 32
 
 devices = [cpu, gpu] = [torch.device('cpu'), torch.device('cuda')]
 
@@ -48,7 +48,7 @@ def _setup(device, levy_area_approximation, shape):
     tb = torch.rand([], device=device)
     ta, tb = min(ta, tb), max(ta, tb)
     bm = torchsde.BrownianInterval(t0=t0, t1=t1, shape=shape, device=device,
-                                   levy_area_approximation=levy_area_approximation)
+                                   levy_area_approximation=levy_area_approximation, pool_size=POOL_SIZE)
     return ta, tb, bm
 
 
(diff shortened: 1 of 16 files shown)


patrick-kidger:
Oh also - I'd like to remove the `.to` method from `BrownianInterval`. (And perhaps for consistency from the others too.)
Is this something we need to support for any reason?

The reason for this is that it turns out that `torch.randn` produces different results on different devices, and for different dtypes [but only provided the shape is large enough, strangely], so the reconstruction procedure used in `BrownianInterval` won't work, unless we always create in float64 on the CPU and copy it across, which I think would be unacceptably slow.

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+@pytest.mark.parametrize("random_order", [False, True])
+def test_continuity(device, levy_area_approximation, random_order):
+    if device == gpu and not torch.cuda.is_available():
+        pytest.skip(msg="CUDA not available.")
+
+    ts = torch.linspace(0., 1., 10000, device=device)
+    bm = torchsde.BrownianInterval(t0=ts[0], t1=ts[-1], shape=(), device=device,
+                                   levy_area_approximation=levy_area_approximation)
+    vals = torch.empty_like(ts)
+    i_ = torch.arange(len(ts), device=device)
```
lxuechen, line 219: Could we rename `i_` and `i` to something more meaningful? Could we also simplify this code for a bit? 
E.g. 
inds = torch.randperm(len(ts), device=device) if random_order else torch.arange(len(ts), device=devce)?

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+    if random_order:
+        i_ = i_[torch.randperm(len(ts), device=device)]
+    for i in i_:
```
lxuechen, line 222: Could we do this with list-comprehension? 
E.g.
vals = [bm(t) for t in ts]?

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+        t = ts[i]
+        vals[i] = bm(t)
+    last_val = vals[0]
+    for val in vals[1:]:
+        assert (val - last_val).abs().max() < 5e-2
+        last_val = val
+
+
+@pytest.mark.skip("BrownianInterval.to not supported")
```
lxuechen, line 231: Why is this test skipped by default?

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+
+    assert w.dtype == w2.dtype == w3.dtype == w4.dtype == torch.float64
+    assert w_.dtype == w2_.dtype == w3_.dtype == w4_.dtype == torch.float32
+    assert torch.allclose(w, w_.double())
+    assert torch.allclose(w2, w2_.double())
+    assert torch.allclose(w3, w3_.double())
+    assert torch.allclose(w4, w4_.double())
+
+
+@pytest.mark.skip("BrownianInterval.to not supported")
```
lxuechen, line 270: Why is this test skipped by default?

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+            rescaled_sample = (sample - mean) / std
+
+            _, pval = kstest(rescaled_sample, 'norm')
+            assert pval >= ALPHA
+
+
+@pytest.mark.parametrize("device", devices)
+@pytest.mark.parametrize("levy_area_approximation", ['none', 'space-time', 'davie', 'foster'])
+@pytest.mark.parametrize("random_order", [False, True])
+def test_continuity(device, levy_area_approximation, random_order):
```
lxuechen, line 211: What is the main purpose of this test?

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+    cache2 = {}
+    for ta_, tb_ in cache:
+        val = bm(ta_, tb_, return_U=return_U, return_A=return_A)
+        if torch.is_tensor(val):
+            val = (val,)
+        cache2[ta_, tb_] = tuple(v.detach().clone() for v in val)
+
+    for ta_, tb_ in cache:
+        for v1, v2 in zip(cache[ta_, tb_], cache2[ta_, tb_]):
+            assert (v1 == v2).all()
```
lxuechen, line 151: Could we do `torch.allclose` since these are floating point tensors?

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+            val0 = (vals[0],)
+        else:
+            val0 = vals[0]
+        for v, v0 in zip(val, val0):
+            assert (v == v0).all()
+
+
+@pytest.mark.parametrize("device", devices)
+@pytest.mark.parametrize("levy_area_approximation, return_U, return_A", _levy_returns())
+def test_determinism_large(device, levy_area_approximation, return_U, return_A):
```
lxuechen, line 127: Could you add a short descriptive docstring for this test, since the function body is a little longer than the others.

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+
+import numpy as np
+import numpy.random as npr
+import torch
+from scipy.stats import norm, kstest
+
+import pytest
+import torchsde
+
+torch.manual_seed(2147483647)
```
lxuechen, line 31: Is there a particular reason to choose 2**31 - 1 as opposed to 0?

tests/test_brownian_tree.py
```diff
@@ -102,6 +102,7 @@ def test_normality(brownian_class, device):
             ref_dist = norm(loc=mean_, scale=std_)
 
             _, pval = kstest(samples_, ref_dist.cdf)
+            print(pval)
```
lxuechen, line 105: Remove plz.

tests/test_brownian_path.py
```diff
@@ -97,3 +97,24 @@ def test_normality(brownian_class, device):
...
+@pytest.mark.parametrize("device", devices)
+@pytest.mark.parametrize("random_order", [False, True])
+def test_continuity(device, random_order):
+    if device == gpu and not torch.cuda.is_available():
+        pytest.skip(msg="CUDA not available.")
+
+    ts = torch.linspace(0., 1., 10000, device=device)
+    bm = torchsde.BrownianPath(t0=ts[0], t1=ts[-1], w0=torch.randn((), dtype=ts.dtype, device=device))
+    vals = torch.empty_like(ts)
+    i_ = torch.arange(len(ts), device=device)
```
lxuechen, line 111: This could also be simplified as described above.

lxuechen:
Thanks for the tests and bug fix. The K-S tests fails on my machine. I think it might be worthwhile to think about the `pool_size` argument for `SeedSequence`. Using a larger pool_size typically helps in normality, but could also drag down the speed.

lxuechen:
> Oh also - I'd like to remove the `.to` method from `BrownianInterval`. (And perhaps for consistency from the others too.)
> Is this something we need to support for any reason?
> 
> The reason for this is that it turns out that `torch.randn` produces different results on different devices, and for different dtypes [but only provided the shape is large enough, strangely], so the reconstruction procedure used in `BrownianInterval` won't work, unless we always create in float64 on the CPU and copy it across, which I think would be unacceptably slow.

Let's remove `to` for BrownianInterval only for now. Thanks for mentioning this.

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,311 @@
...
+    vals = torch.empty_like(ts)
+    i_ = torch.arange(len(ts), device=device)
+    if random_order:
+        i_ = i_[torch.randperm(len(ts), device=device)]
+    for i in i_:
+        t = ts[i]
+        vals[i] = bm(t)
+    last_val = vals[0]
+    for val in vals[1:]:
+        assert (val - last_val).abs().max() < 5e-2
```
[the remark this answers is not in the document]
patrick-kidger: It's the value that worked for `BrownianPath`.

lxuechen:
> I don't actually follow what you were saying about there being a bug in ReverseBrownian?

I fixed this on `dev-strat-adjoint` and added a docstring. Ultimately, it also has to do with how the adjoint functions are negated. I'm inclined to keep the current form for the adjoints.

tests/test_brownian_interval.py
patrick-kidger: Done.

tests/test_brownian_interval.py
patrick-kidger: No, because the insertions into `val` are in general out of order.

tests/test_brownian_interval.py
patrick-kidger: Because of the issues with `BrownianInterval.to` as below.

tests/test_brownian_interval.py
patrick-kidger: Because of the issues with `BrownianInterval.to` as below.

tests/test_brownian_interval.py
patrick-kidger: As the name suggests, to test continuity ;)

I found (pre-bugfixes) that `BrownianInterval` would sometimes produce discontinuous paths, so this felt like a sensible sanity check for that issue.

tests/test_brownian_interval.py
patrick-kidger: That would only test closeness; here I specifically want to test for equality, i.e. perfect determinism.

tests/test_brownian_interval.py
patrick-kidger: Sure.

tests/test_brownian_interval.py
patrick-kidger: Whoops, didn't mean to choose a magic number like that. (I copied and pasted it from the PyTorch docs without thinking.)

Whilst obviously this particular choice doesn't help, it was meant in response to this line from the PyTorch docs re: random seeds:
"""
It is recommended to set a large seed, i.e. a number that has a good balance of 0 and 1 bits. Avoid having many 0 bits in the seed.
"""
I don't know how much that would actually affect us, but I thought it best to follow good practice.

tests/test_brownian_tree.py
patrick-kidger: Whoops!

tests/test_brownian_path.py
patrick-kidger: Yup.

tests/test_brownian_interval.py
lxuechen: I agree with this now.

tests/test_brownian_interval.py
lxuechen: SGTM.

tests/test_brownian_interval.py
lxuechen: SGTM.

lxuechen:
As a minor side note, could we give the branches more meaningful names. It'd be nice if we could try to dedicate each branch to a single PR and remove the branch afterwards, so as not to produce more and more branches?

patrick-kidger:
> 
> 
> Thanks for the tests and bug fix. The K-S tests fails on my machine. I think it might be worthwhile to think about the `pool_size` argument for `SeedSequence`. Using a larger pool_size typically helps in normality, but could also drag down the speed.

Increased (and added an argument) for the pool size. I've set it to 8 by default; as the queries in the BrownianInterval are much shallower than the BrownianTree (which uses pool size 24) then I think this should be fine. (And the KS test now passes; I forgot the seed would make that deterministic.)

> 
> 
> > I don't actually follow what you were saying about there being a bug in ReverseBrownian?
> 
> I fixed this on `dev-strat-adjoint` and added a docstring; see [this](https://github.com/google-research/torchsde/blob/2171e23a08f923319fdb11a31017b263493f7699/torchsde/_brownian/modified.py#L50). Ultimately, it also has to do with how the adjoint functions are negated. I'm inclined to keep the current form for the adjoints.

Makes sense.

> 
> 
> As a minor side note, could we give the branches more meaningful names. It'd be nice if we could try to dedicate each branch to a single PR and remove the branch afterwards, so as not to produce more and more branches?

Aha sorry. I usually tend to end up adding a mix of interdependent things and wanting to rename the branch afterwards, so I've ended up just picking deliberately not meaningful names. Can definitely remove branches afterwards.

tests/test_brownian_interval.py
lxuechen: That's interesting, do you have any guesses as to why it would produce discontinuous paths?

tests/test_brownian_interval.py
lxuechen: If we were to remove the `to` function, could we also remove the corresponding tests?

tests/test_brownian_interval.py
lxuechen: Same here.

tests/test_brownian_interval.py
patrick-kidger: Depends - do we want to leave the possibility of supporting this open in the future (in which case let's leave it in but skipped), or do we want to remove it entirely? No strong feelings either way.

lxuechen: If we leave it here, could we at least put a todo?

tests/test_brownian_interval.py
```diff
@@ -0,0 +1,315 @@
...
+    if device == gpu and not torch.cuda.is_available():
+        pytest.skip(msg="CUDA not available.")
+
+    t0_, t1_ = 0.0, 1.0
+    t0, t1 = torch.tensor([t0_, t1_], device=device)
+    for _ in range(REPS):
+        W = torch.tensor(npr.randn(), device=device).repeat(LARGE_BATCH_SIZE)
+        bm = torchsde.BrownianInterval(t0=t0, t1=t1, W=W, levy_area_approximation=levy_area_approximation)
+
+        for _ in range(REPS):
```
lxuechen, line 172: Have you run this test? It looks somewhat suspicious to me that it will pass. As you repeat the bm query, the conditional statistics would become different than just conditioning on the left most and right most endpoints (t0, t1 in this case), right?

tests/test_brownian_interval.py
patrick-kidger: Sure.

tests/test_brownian_interval.py
patrick-kidger: Yes, I've run it. It should work in terms of theory.
If I query the BM and then ignore my observation then I might as well have not queried it at all - distributionally (or in this case, over a batch of size 32000), it's the same.

If you prefer, this is tower law. For any one sample, sure, the conditional statistics change. Take the expectation (over a batch of size 32000) and the condition drops out.

lxuechen: That makes sense, but I as run your branch on my local machine, these normality tests fail specifically. Am I missing anything?

lxuechen: I'm getting pvals very much close to being 0.

patrick-kidger: Ah, interestingly it looks like there's some nondeterminism involved. Sometimes I get all passes; sometimes I get a single failure mixed in as well. I'm not sure where that's coming from. (Sidenote: the GPU tests were never going to be deterministic because of nondeterminism in CUDA / nonassociativity of floating point arithmetic, but I'm also observing this on the CPU as well.)

Regardless, the pvals I observe seem pretty good; here's a sample from one run:
```
0.1831880731221287
0.7203067705273924
0.4005865963923665
0.9836020139630008
0.595396068278881
0.754736013617661
0.49089814175274626
0.8991110938461623
0.9471682808052452
0.06838178984312837
0.1917608973567274
0.10887317980591127
0.4471597979510218
0.2429775832482548
0.5234704439252269
0.9927171941269395
0.7041367966736902
0.6697926636524723
0.966971540959411
0.31859260562945524
0.49846075734446305
0.9633103124760822
0.25049223541296295
0.7333669455911738
0.8127956811078841
0.7822939021877175
0.8943048913302614
0.655253455597571
0.7993983708378245
0.14585232647852497
0.15457143178071975
0.8382653075582929
0.6702470486709494
0.4400739985187604
0.7388889139761349
0.10928069948262087
0.2908813124306524
0.12164793337455271
0.2304577968211593
0.6821789375137386
0.3738747784267529
0.15103241471340767
0.09241350508792956
0.13738289156372208
0.9355751354566113
0.19981707556298756
0.17567687104752305
0.875303717188714
0.6463672817110033
0.7680254937059569
0.9855047039836732
0.9096026698627325
0.47539405333075624
0.0021910005327920225
0.14046700256713318
0.31588096252673215
0.6771577262131989
0.9723814652857421
0.036046821301315646
0.3272079086543058
0.3236566425640489
0.4085607048407861
0.030045800063961514
0.8577012449476132
0.6006252116500315
0.33018822367454104
0.6688857325347278
0.8692185469427525
0.006884728052989697
0.09794983997900208
0.9295964081955631
```
I wouldn't be rejecting the null hypothesis as a result of these.

Can you double check you've got the latest version of the code, and share what kind of p-values you observe?

lxuechen: Yes, I was running the on the latest commit. Here're some pvals I get from running the conditional normal test
```
0.5607581684785437
0.5919312036699521
0.014969262467999029
0.9024321662628461
0.48012136689587176
5.224794616863965e-78
0.5453968496808617
4.074461892756811e-20
2.121709516398075e-181
```

lxuechen: This is on CPU btw.

patrick-kidger: That's fascinating. Perhaps whatever PRNG is being used under the hood in PyTorch is different between us, and for whatever reason mine of better quality?
Can you try re-running on a different machine, preferably a different OS/architecture/etc. _(you know things are getting interesting when that's a suggestion)_ and see what happens?
Additionally, also try increasing the pool size and see if that changes things?

lxuechen: Yeah I agree this is interesting. Switching machines and increasing pool size did not help. The tests are failing for the following environments for me
```
PyTorch version: 1.6.0
Is debug build: False
CUDA used to build PyTorch: None

OS: Mac OSX 10.15.4 (x86_64)
GCC version: Could not collect
Clang version: 11.0.3 (clang-1103.0.32.29)
CMake version: version 3.16.5

Python version: 3.7 (64-bit runtime)
Is CUDA available: False
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
```

```
PyTorch version: 1.6.0+cu101
Is debug build: False
CUDA used to build PyTorch: 10.1

OS: Debian GNU/Linux 9.12 (stretch) (x86_64)
GCC version: (Debian 6.3.0-18+deb9u1) 6.3.0 20170516
Clang version: Could not collect
CMake version: Could not collect

Python version: 3.7 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: 10.1.243
GPU models and configuration:
GPU 0: Tesla P100-PCIE-16GB
GPU 1: Tesla P100-PCIE-16GB
GPU 2: Tesla P100-PCIE-16GB
GPU 3: Tesla P100-PCIE-16GB

Nvidia driver version: 418.87.01
cuDNN version: Could not collect
```
Did you push the latest commits to your branch? Also could you share your system info?

lxuechen: I guess in the short term, just to ensure the tests aren't too flaky, we should probably reduce REPS/increase batch size or altogether change/remove the tests here. 

Definitely worth coming back for another revisit.

patrick-kidger: Yep, definitely using the latest version.

Things seem fine for me with either Python 3.7 or Python 3.8 on Linux.

What versions of the relevant libraries do you have? I could believe that a difference in scipy or mkl_random could affect things.

My environments:

Environment 1:
```
PyTorch version: 1.6.0
Is debug build: False
CUDA used to build PyTorch: 10.2

OS: Ubuntu 18.04.5 LTS (x86_64)
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Clang version: 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
CMake version: version 3.10.2

Python version: 3.7 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration:
GPU 0: GeForce RTX 2080 Ti
GPU 1: GeForce RTX 2080 Ti
GPU 2: Quadro GP100
GPU 3: Quadro GP100

Nvidia driver version: 450.51.06
cuDNN version: /usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5

Versions of relevant libraries:
[pip3] numpy==1.19.1
[pip3] torch==1.6.0
[pip3] torchsde==0.1.2
[conda] blas                      1.0                         mkl
[conda] cudatoolkit               10.2.89              hfd86e86_1
[conda] mkl                       2020.1                      217
[conda] mkl-service               2.3.0            py37he904b0f_0
[conda] mkl_fft                   1.1.0            py37h23d657b_0
[conda] mkl_random                1.1.1            py37h0573a6f_0
[conda] numpy                     1.19.1           py37hbc911f0_0
[conda] numpy-base                1.19.1           py37hfa32c7d_0
[conda] pytorch                   1.6.0           py3.7_cuda10.2.89_cudnn7.6.5_0    pytorch
[conda] torchsde                  0.1.2                     dev_0    <develop>
```

Environment 2:
```
PyTorch version: 1.6.0
Is debug build: False
CUDA used to build PyTorch: 10.2

OS: Ubuntu 18.04.5 LTS (x86_64)
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Clang version: 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
CMake version: version 3.10.2

Python version: 3.8 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration:
GPU 0: GeForce RTX 2080 Ti
GPU 1: GeForce RTX 2080 Ti
GPU 2: Quadro GP100
GPU 3: Quadro GP100

Nvidia driver version: 450.51.06
cuDNN version: /usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5

Versions of relevant libraries:
[pip3] numpy==1.19.1
[pip3] torch==1.6.0
[pip3] torchsde==0.1.2
[conda] blas                      1.0                         mkl
[conda] cudatoolkit               10.2.89              hfd86e86_1
[conda] mkl                       2020.1                      217
[conda] mkl-service               2.3.0            py38he904b0f_0
[conda] mkl_fft                   1.1.0            py38h23d657b_0
[conda] mkl_random                1.1.1            py38h0573a6f_0
[conda] numpy                     1.19.1           py38hbc911f0_0
[conda] numpy-base                1.19.1           py38hfa32c7d_0
[conda] pytorch                   1.6.0           py3.8_cuda10.2.89_cudnn7.6.5_0    pytorch
[conda] torchsde                  0.1.2                     dev_0    <develop>
```

patrick-kidger: Regarding flakiness of tests - do you want to make the adjustments for that, as I have no way of testing what makes it not flakey! I'll leave this branch alone for you to make changes on.

Unrelatedly, is there anything you think I should particularly look at next? There's bits and bobs I can do but I think I've handled all the glaring items on my to-do list.

patrick-kidger: Oh, and it seems like I didn't even list my own version of scipy! For 3.7, 3.8 respectively:
```
scipy                     1.5.0            py37h0b6359f_0
scipy                     1.5.0            py38h0b6359f_0
```

lxuechen: I upgraded all the packages to have the same versions as you do, and the problem didn't go away. 

Could you double check if you get mean=0 and std=1 for the test [here](https://github.com/google-research/torchsde/blob/cae4bc2dd9b4ab41be8a3788dccba3437915c0e3/tests/test_brownian_interval.py#L186)? 

I'm seeing things fail systematically for `foster`/`david`/`space-time`.

I'm sorry if this gets grueling, but if things fail here or aren't stable enough, I'm pretty sure they'll fail even more miserably when we plug everything into the sdeint_adjoint code, and frankly debugging a `torch.nn.Function` can get annoying.

lxuechen: A side note is that when I was testing bpath and btree, I noticed that things can get flaky when the time points are very close, so that's why I added an `eps` guard to prevent that from happening.

torchsde/_brownian/brownian_interval.py
```diff
@@ -378,12 +370,21 @@ def __init__(self,
...
+        if shape is not None:
+            shapes.append(shape)
+        if dtype is None:
+            if len(dtypes) == 0:
+                dtypes.append(torch.get_default_dtype())
+        else:
+            dtypes.append(dtype)
+        if device is None:
+            if len(devices) == 0:
+                # Is there no better way to get the current default device?
```
lxuechen, line 382: Why not just use CPU?

patrick-kidger: So I had it in my head that there was some way to set a non-CPU default device, but I think you're right that there isn't. I'll change this.

tests/test_brownian_interval.py
patrick-kidger: > 
> 
> I upgraded all the packages to have the same versions as you do, and the problem didn't go away.
> 
> Could you double check if you get mean=0 and std=1 for the test [here](https://github.com/google-research/torchsde/blob/cae4bc2dd9b4ab41be8a3788dccba3437915c0e3/tests/test_brownian_interval.py#L186)?
> 
> I'm seeing things fail systematically for `foster`/`david`/`space-time`.
> 
> I'm sorry if this gets grueling, but if things fail here or aren't stable enough, I'm pretty sure they'll fail even more miserably when we plug everything into the sdeint_adjoint code, and frankly debugging a `torch.nn.Function` can get annoying.

Adding `print(torch.std_mean(torch.tensor(rescaled_sample)))` to the test prints out:
```
(tensor(1.0025), tensor(0.0006))
(tensor(1.0014), tensor(-0.0026))
(tensor(1.0000), tensor(-0.0006))
(tensor(1.0020), tensor(-0.0018))
(tensor(1.0009), tensor(-0.0034))
(tensor(1.0010), tensor(0.0022))
(tensor(0.9997), tensor(0.0007))
(tensor(0.9977), tensor(0.0029))
(tensor(0.9999), tensor(0.0016))
(tensor(0.9971), tensor(0.0017))
(tensor(0.9979), tensor(0.0020))
(tensor(0.9994), tensor(-0.0053))
(tensor(1.0013), tensor(-0.0019))
(tensor(1.0012), tensor(0.0018))
(tensor(0.9993), tensor(-0.0008))
(tensor(1.0021), tensor(-0.0033))
(tensor(0.9992), tensor(-0.0011))
(tensor(1.0021), tensor(0.0064))
(tensor(1.0039), tensor(-0.0108))
(tensor(0.9979), tensor(0.0018))
(tensor(0.9984), tensor(-0.0030))
(tensor(0.9970), tensor(0.0034))
(tensor(0.9989), tensor(-0.0063))
```
which LGTM.

And don't worry, I appreciate that "it works on my machine" isn't going to cut it.

Are you saying that `none` seems to work? (That uses a different algorithm to the other three, so I can believe it.)

Okay, to keep digging in: when you run
```

Comment (patrick-kidger) [reply tests/test_brownian_interval.py]: > 
> 
> A side note is that when I was testing bpath and btree, I noticed that things can get flaky when the time points are very close, so that's why I added an `eps` guard to prevent that from happening.

What kind of flakiness do you observe? I didn't _too_ much thought into numerical stability whilst I was writing things (although I did a bit). If possible I'd rather rewrite the computations if they're unstable rather than add another feature to the mix.

**Comment by patrick-kidger** [comment]
@lxuechen Needing to merge this is holding things up. (In particular the fact that this fixes calling `g_prod` and therefore being able to use `sdeint`!) Can we merge it and open the normality discussion as an issue?

File: tests/test_brownian_interval.py (hunk 12 above)
Comment (lxuechen) [reply tests/test_brownian_interval.py]: > do you get 2029077620 as the valueI believe SeedSequence is a fixed implementation that should be the same for both of us, so it's a good place to start.

I get 
```771466205```

> torch.randn(64, generator=torch.Generator().manual_seed(seed))

I get 
```
tensor([ 0.7314, -0.7575,  0.5196, -0.2246, -1.1200,  1.2430,  1.2759,  0.0239,
        -0.1964, -0.4379, -0.8360, -0.7990, -0.7343,  0.7382,  2.3253, -0.0809,
        -1.3725, -0.4227, -0.3373, -1.7258, -0.8709,  0.1395,  1.2477, -0.1724,
         0.3683, -0.7075,  1.1928, -0.2962,  0.8203, -1.7916, -0.7161,  0.1629,
        -0.3906,  0.3133,  0.0258, -0.3784,  0.3754, -0.3247,  1.1662, -0.7479,
        -1.9184,  1.1434,  1.7734, -0.0771,  1.1552, -0.4181,  0.6999,  0.5334,
         0.7735,  0.2979, -1.7747, -0.4775,  0.2538,  0.2146, -1.2433,  1.2250,
         0.6106, -2.3741,  0.6630, -1.3142,  0.8098,  0.8615, -1.0300, -1.7857])
```

On a broader note, I don't think these check are indicative of much. 

> What kind of flakiness do you observe? I didn't too much thought into numerical stability whilst I was writing things (although I did a bit). If possible I'd rather rewrite the computations if they're unstable rather than add another feature to the mix.

So when the query time is very close to one side for Bbridge, there could be a problem of adding large numbers to small numbers which is not so great in the context of floating point computation.

Comment (lxuechen) [reply tests/test_brownian_interval.py]: Cloning a fresh repo and reinstalling your branch, with `print(torch.std_mean(torch.tensor(rescaled_sample)))`, I'm getting

```
(tensor(0.9987), tensor(0.0003))
(tensor(0.9980), tensor(-0.0024))
(tensor(0.9973), tensor(0.0036))
(tensor(0.6418), tensor(-0.0011))
(tensor(0.6476), tensor(0.0030))
(tensor(0.8856), tensor(5.2828e-05))
```

lxuechen:
> @lxuechen Needing to merge this is holding things up. (In particular the fact that this fixes calling `g_prod` and therefore being able to use `sdeint`!) Can we merge it and open the normality discussion as an issue?

I can merge this after sending another minor commit, but let's keep discussing the problem here, instead of opening up a new issue.

tests/test_brownian_interval.py
patrick-kidger: Au contraire, I think it's identified the exact problem. You're using an old version of numpy; upgrade to the latest 1.19.1.

lxuechen: OK, it's working now. Could you send in another PR bumping up the version in `setup.py`?

PR #33, opened 2020-08-19
Tweaked to not hang on adaptive solvers.

patrick-kidger:
From the discussion in #28. Figured out what was going on and fixed it.

I can go into details if you're curious.

### DIFF
--- a/torchsde/_core/base_solver.py
+++ b/torchsde/_core/base_solver.py
@@ -76,8 +76,9 @@ class BaseSDESolver(metaclass=better_abc.ABCMeta):
         """
         raise NotImplementedError
 
-    def step_logqp(self, t, y, dt, logqp0):
-        t1, y1 = self.step(t, y, dt)
+    def step_logqp(self, t, next_t, y, logqp0):
+        dt = next_t - t
+        y1 = self.step_(t, next_t, y)
 
         if self.sde.noise_type in ("diagonal", "scalar"):
             f_eval = self.sde.f(t, y)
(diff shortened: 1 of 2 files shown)


lxuechen:
Thanks for the fix! Could you describe a little more of what you think the problem is? 

Also, I think we should merge #21 first as it was completed last night, and also merge #31 before this.

patrick-kidger:
Sure thing. Previously we had a setup like:
```
while a < b:
    step = ...
    step = min(step, b - a)
    ...
    a += step
```
eventually `a` gets close enough to `b` that the `min` kicks in and `step = b - a`.
You'd expect to then get `a += step` implying that `a <- a + (b - a) = b` and the while loop to complete. I think what actually happens is that `a + (b - a) < b` due to rounding errors, so the loop never terminates. An explicit example for which this condition is possible: `a = 1e30`, `b = 1`. Then the LHS is 0 but the RHS is 1.
[Obviously this example can't be exactly what's going on because we additionally have `a < b` in our case, but I expect it's something like this.]

The trick is just not to work with delta-ts; instead explicitly work with the points that you're stepping between.

And ah, sorry I didn't see that #21 is done - I think I might not get notifications from mentions in edits? (Idk.) I'll start reviewing it.

lxuechen:
Since #21 is merged, I think it'd be nice for us to have this tweak done more completely. In particular, I'd recommend changing the step function for the solvers to be `def step(self, t0, y0, t1)`. Alternatively, we could stick with the wrapper step function, but I think we should have a wrapper for both `step` and `step_logqp` just to mirror the functionality. This might help me in the future when I fix the logqp component.

patrick-kidger:
Changing all the solvers' interfaces is what I'm thinking too. I'm hoping to get this PR done tomorrow.

PR #34, opened 2020-08-20
Added BrownianReturn

patrick-kidger:
Just a quick PR on what I think is overlapping functionality for both of us, to make sure we don't cross wires on this later.


lxuechen:
LGTM.

Issue #36, opened 2020-08-23
[On dev branch] New BrownianInterval hangs when `dt` is very small


lxuechen:
I think this relates to the issue @patrick-kidger mentioned that when running the scalar diagnostics, large amounts of memory is consumed. I'm seeing this also on my machine. 

I've managed to pinpoint the issue: When `tb-ta` is very small, the dependency tree creation phase hangs, with `_set_points` being repeated indefinitely. More specifically, [this](https://github.com/google-research/torchsde/blob/4c70f3e0207a8a9b27f2a91ccc0b4e8cda3fd431/torchsde/_brownian/brownian_interval.py#L539) line.

patrick-kidger:
How small is small? I've tested this down to `dt=1e-7` (so 100 times smaller than the default `dt_min`) and things seem fine. (It terminates in 80 seconds on my machine.)

lxuechen:
Good point, I'm observing values of `tb - ta` to be as small as `1.7763568394002505e-15`. This is really a result of the while loop in `base_solver.py` still stepping forward despite `curr_t` being set to ts[-1].

This is clearly due to limited floating point precision, though the precise reason I have not figured out yet.

patrick-kidger:
Are you saying there's still a bug in `base_solver.py` somewhere?

This has made me realise, though that we can get an arbitrarily small timestep on the final step. I'll update the heuristics for creating the dependency tree to be a bit smarter.

lxuechen:
That is clearly the case, i.e. curr_t = ts[-1] < out_t=ts[-1].

patrick-kidger:
Can you provide a repro? Just eyeballing the integration code nothing really looks wrong to me.

lxuechen:
> Can you provide a repro? Just eyeballing the integration code nothing really looks wrong to me.

Try running `python -m diagnostics.ito_scalar` on branch `dev-remove-tuple`. The problem occurs regardless of which random seed I use on my machine. 

I think ultimately, we should step pass the boundary and pull back later on.

patrick-kidger:
Right, figured it out - this was a bug in BInterval, not in the solver. I'll have a PR in a few moments.

Issue #27, opened 2020-08-17
Scalar noise SDE shape check fails


lxuechen:
Latest on dev, the contract checking fails for the following
```python
import torch

from torchsde import BrownianInterval, BaseSDE, sdeint

device = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')

batch_size, d, m = 16, 3, 2
t0, t1 = 0.0, 0.3
dtype = torch.get_default_dtype()
y0 = torch.zeros(batch_size, d, device=device)


class ScalarSDE(BaseSDE):

    def f(self, t, y):
        return torch.sin(y)

    def g(self, t, y):
        return torch.cos(y).sigmoid()


sde = ScalarSDE(sde_type='ito', noise_type='scalar')
bm = BrownianInterval(t0=t0, t1=t1, shape=(batch_size, 1), dtype=dtype, device=device, levy_area_approximation='foster')
sdeint(sde=sde, y0=y0, ts=[t0, t1], bm=bm)
```

Also it might be worthwhile to characterize exactly what API we'd like to use. I think we should go with `g` outputting `(batch_size, d, 1)` or `(batch_size, d)` or preferably both if possible for backward compatibility.

patrick-kidger:
Yes, this was a concious choice. At the moment it checks for shape `(..., d, 1)`.

This raises an interesting point: I think it's reasonable to support either multiple batch dimensions [and I realise much of the rest of the code doesn't yet do this], or to support scalar diffusion having either of the shapes you mention, but not both. The logic for what counts as batch and what doesn't would end up being rather strange, then.

Personally I would lean quite strongly in favour of supporting multiple batch dimensions. I made the concious decision not to in previous projects, and have since regretted it.

PR #37, opened 2020-08-23
[On dev branch] Tuple rewrite

lxuechen:
Contents in this PR:
- Remove support for tuple `y0` and tuple-based `sde` for `sdeint`.
  - Rationale: This feature is somewhat niche. Supporting this would likely require adding an alternative tuple-based `ForwardSDE` that has the correct versions of `f`, `g_prod`, `gdg_prod`, and `dg_ga_jvp` (per our email discussion). 
- Make `AdjointSDE` take in `ForwardSDE` and use the latter's `g_prod`.
  - This simplifies `g_prod` of the former.
- Make `AdjointSDE` take in (and output) flattened tensors and perform flattening/unflattening inside the registered functions. 
  - One advantage of this is that we only need to unflatten 2 tensors now. 
- Refactor solvers to not be based on tuples.
- Explicit flattening/unflattening in `adjoint.py`
  - This differs slightly from the current implementation in `torchdiffeq`, where the `_TupleFunc` construction is in `_check_inputs`.
  - Rationale: We assume that `sdeint.integrate` only supports tensor-based `y0` and `sde`, and we should enforce this usage for now. 
- Filter out the parameters that don't require gradients in `sdeint_adjoint`.
- Fix plot folder names for files in `diagnostics`.
- *Temporarily* deprecating `logqp` support. 
  - This should either be added back or the example latent_sde rewritten with state augmentation when we get the general noise adjoint working with the log-ODE scheme. (I view getting the log-ODE scheme-based adjoint working for general noise Strat SDE as a milestone. )
- Fix scalar noise shape check and it's functionality in `sdeint` inconsistency. Hence, resolving #27.

The rate diagnostics files run, but I haven't been able to obtain rates due to #36. I need to run these when this is fixed. All tests pass.

Overall, this PR should also make adding `dg_ga_jvp` for the adjoint easier.


tests/test_strat.py
```diff
@@ -75,38 +75,35 @@ def _batch_jacobian(output, input_):
     return torch.stack(jacs, dim=0)
 
 
-def _gdg_jvp_brute_force(sde, t, y, a):
+def _dg_ga_jvp_brute_force(sde, t, y, a):
     # Only returns the value for the first input-output pair.
```
patrick-kidger, line 79: Is this referring to the re-rewrite potential of having tuples? If so it can be removed.

torchsde/_core/adjoint.py
```diff
@@ -167,24 +70,16 @@ def backward(ctx, *grad_outputs):
         dt_min = ctx.dt_min
         adjoint_options = ctx.adjoint_options
 
-        params = misc.make_seq_requires_grad(sde.parameters())
-        n_tensors, n_params = len(ans), len(params)
-
+        aug_state = (ys[-1], grad_ys[-1], *[torch.zeros_like(param) for param in params])
```
patrick-kidger, line 73: Nitpick: I think this would be slightly neater as `(ys[-1], grad_ys[-1]) + tuple(torch.zeros_like(param) for param in params)`.

torchsde/_core/adjoint.py
```diff
@@ -195,27 +90,22 @@ def backward(ctx, *grad_outputs):
                 dt_min=dt_min,
                 options=adjoint_options
             )
+            aug_state = misc.flat_to_shape(aug_state[1], shapes)  # Unpack the state at time -ts[i - 1].
+            aug_state[0] = ys[i - 1]
+            aug_state[1] += grad_ys[i - 1]
```
patrick-kidger, line 95: In principle is `sdeint_adjoint` going to support double backward? I've been thinking that this may actually be a very important thing for us to support wrt our current project.
(I hadn't mentioned it as I didn't want to add to your workload. But here everything looks like it should pretty much work. At least in principle; not sure how accurate the double-gradients will actually be.)

I'd suggest changing this to an out-of-place update just in case, and changing the call to `sdeint.integrate` to `_SdeintAdjointMethod.apply`, and I think we should be good to go.

torchsde/_core/base_sde.py
```diff
@@ -48,117 +48,116 @@ def __init__(self, sde):
         if hasattr(sde, "h"):
```
patrick-kidger, line 48: This can be removed now.

torchsde/_core/base_sde.py
```diff
@@ -48,117 +48,116 @@ def __init__(self, sde):
...
+            dg_ga_jvp = dg_ga_jvp.diagonal(dim1=-2, dim2=-1).sum(-1)
+        return dg_ga_jvp
 
-    def h(self, t, y):
-        return self._base_sde.h(t, y[0]),
+    def _return_zero(self, t, y, v):  # noqa
+        return 0.
 
 
 class RenameMethodsSDE(BaseSDE):
```
patrick-kidger, line 163: This doesn't need `h` any more.

torchsde/_core/methods/milstein.py
```diff
@@ -50,55 +49,47 @@ def v_term(self, I_k, dt):
...
-        g_prod_eval = self.sde.g_prod(t0, y0, I_k)
+        f = self.sde.f(t0, y0)
+        g_prod_I_k = self.sde.g_prod(t0, y0, I_k)
 
         if self.options[METHOD_OPTIONS.grad_free]:
-            g_eval = self.sde.g(t0, y0)
-            g_prod_eval_v = self.sde.g_prod(t0, y0, v)
+            g = self.sde.g(t0, y0)
+            g_prod_v = self.sde.g_prod(t0, y0, v)
             sqrt_dt = torch.sqrt(dt) if isinstance(dt, torch.Tensor) else math.sqrt(dt)
```
patrick-kidger, line 66: I think `dt` can only be a tensor now (as input times are now coerced to tensor), so this can be simplified.

torchsde/_core/methods/srk.py
```diff
@@ -63,38 +59,31 @@ def diagonal_or_scalar_step(self, t0, t1, y0):
         dt = t1 - t0
         sqrt_dt = torch.sqrt(dt) if isinstance(dt, torch.Tensor) else math.sqrt(dt)
         I_k, I_k0 = self.bm(t0, t1, return_U=True)
-        I_kk = [(delta_bm_ ** 2. - dt) / 2. for delta_bm_ in I_k]
-        I_kkk = [(delta_bm_ ** 3. - 3. * dt * delta_bm_) / 6. for delta_bm_ in I_k]
+        I_kk = (I_k ** 2 - dt) / 2.
+        I_kkk = (I_k ** 3 - 3. * dt * I_k) / 6.
```
patrick-kidger, line 63: Can these divisions be substituted for multiplications? (Making `_r6 = 1/6` a global constant is a little cleaner; writing out `0.16666...67` to full precision is a little faster; IMO doesn't really matter which is done.)

torchsde/_core/methods/srk.py
```diff
@@ -63,38 +59,31 @@ def diagonal_or_scalar_step(self, t0, t1, y0):
...
-                    H0s_ + srid2.A0[s][j] * f_eval_ * dt + srid2.B0[s][j] * g_eval_ * I_k0_ / dt
-                    for H0s_, f_eval_, g_eval_, I_k0_ in zip(H0s, f_eval, g_eval, I_k0)
-                ]
-                H1s = [
-                    H1s_ + srid2.A1[s][j] * f_eval_ * dt + srid2.B1[s][j] * g_eval_ * sqrt_dt
-                    for H1s_, f_eval_, g_eval_ in zip(H1s, f_eval, g_eval)
-                ]
+                f = self.sde.f(t0 + srid2.C0[j] * dt, H0[j])
+                g = self.sde.g(t0 + srid2.C1[j] * dt, H1[j])
+                g = g.squeeze(2) if g.dim() == 3 else g
```
patrick-kidger, line 72: This is going to break if/when we support multiple batch dimensions. (Plus it's a bit opaque what it's doing.) Factor this out to a different method for additive/diagonal a la Milstein?

torchsde/_core/misc.py
```diff
@@ -109,12 +115,26 @@ def grad(outputs, inputs, **kwargs):
...
+
+def flat_to_shape(tensor, shapes, length=()):
+    tensor_list = []
+    total = 0
+    for shape in shapes:
+        next_total = total + shape.numel()
+        # It's important that this be view((...)), not view(...). Else when length=(), shape=() it fails.
+        tensor_list.append(tensor[..., total:next_total].view((*length, *shape)))
+        total = next_total
+    return tensor_list
```
patrick-kidger, line 140: There's now several unused functions in this file, that can be removed.

torchsde/_core/sdeint.py
```diff
@@ -17,44 +17,41 @@
...
-from .._brownian import BaseBrownian, TupleBrownian, BrownianInterval
+from . import misc
+from .._brownian import BaseBrownian, BrownianInterval
 from ..settings import SDE_TYPES, NOISE_TYPES, METHODS, LEVY_AREA_APPROXIMATIONS
 from ..types import TensorOrTensors, Scalar, Vector
 
 
-def sdeint(sde: [base_sde.BaseSDE],
+def sdeint(sde: base_sde.BaseSDE,
            y0: TensorOrTensors,
```
patrick-kidger, line 29: This should just be a tensor now. I've not checked the whole codebase exhaustively, but do we need TensorOrTensors any more?

torchsde/_core/sdeint.py
```diff
@@ -17,44 +17,41 @@
...
+           method: Optional[str] = "srk",
            dt: Optional[Scalar] = 1e-3,
            adaptive: Optional[bool] = False,
            rtol: Optional[float] = 1e-5,
            atol: Optional[float] = 1e-4,
            dt_min: Optional[Scalar] = 1e-5,
            options: Optional[Dict[str, Any]] = None,
-           names: Optional[Dict[str, str]] = None) -> TensorOrTensors:
+           names: Optional[Dict[str, str]] = None,
+           **unused_kwargs) -> TensorOrTensors:
```
patrick-kidger, line 40: Returns just a tensor now.

torchsde/_core/sdeint.py
```diff
@@ -93,134 +89,108 @@ def sdeint(sde: [base_sde.BaseSDE],
...
-    return results
 
 
-def check_contract(sde, y0, ts, bm, logqp, method, names):
+def check_contract(sde, y0, ts, bm, method, names):
     if names is None:
         names_to_change = {}
     else:
-        names_to_change = {key: names[key] for key in ('drift', 'diffusion', 'prior_drift') if key in names}
+        names_to_change = {key: names[key] for key in ("drift", "diffusion", "prior_drift") if key in names}
```
patrick-kidger, line 99: No more prior drift.

patrick-kidger:
Also, ` tests/__init__.pyc` shouldn't be included in the commit.

patrick-kidger:
Overall, things are looking so much cleaner now. These changes make me very happy.

tests/basic_sde.py
```diff
@@ -129,11 +129,11 @@ def h(self, t, y):
...
-        self.g_param = nn.Parameter(torch.sigmoid(torch.randn(1, d)), requires_grad=True)
+        self.g_param = nn.Parameter(torch.sigmoid(torch.randn(1, d, 1)), requires_grad=True)
         self.noise_type = "scalar"
 
     def g(self, t, y):
-        return self.g_param.repeat(y.size(0), 1)
+        return self.g_param.repeat(y.size(0), 1, 1)
 
 
 class TupleSDE(SDEIto):
```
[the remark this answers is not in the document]
lxuechen: Done.

tests/test_strat.py
lxuechen: No, this is simply a rename, i.e. `gdg_jvp` -> `dg_ga_jvp`. I think the latter is clearer, since it matches the math better.

The rationale of including this function is for testing purposes, i.e. we have more confidence in the two variants of jvp I implemented, and it would come in handy in the future if we ever decide to modify any aspect of those functions.

torchsde/_core/adjoint.py
lxuechen: Sure. I'll be using lists for the second part now, as list comprehension is slightly faster.

torchsde/_core/base_sde.py
lxuechen: Done.

tests/test_strat.py
patrick-kidger: I mean the comment, not the name!

torchsde/_core/adjoint.py
patrick-kidger: Interesting, I didn't know listcomps were faster. Ty.

torchsde/_core/sdeint.py
lxuechen: Done.

torchsde/_core/sdeint.py
lxuechen: Done.

torchsde/_core/misc.py
lxuechen: Done.

torchsde/_core/methods/srk.py
lxuechen: I think we should just write wrappers at some point collapsing/uncollapsing the extra dimensions, since writing a single wrapper class is less work than changing all the solvers.

torchsde/_core/methods/srk.py
lxuechen: Done.

torchsde/_core/methods/milstein.py
lxuechen: Done.

torchsde/_core/adjoint.py
lxuechen: I can change to out-of-place update for now. Though, I'm not inclined to switch from `sdeint.integrate` to `_SdeintAdjointMethod.apply` for now. 

I like testing :), and I think we should do that in another PR after testing that nothing breaks just to be safe.

tests/test_strat.py
lxuechen: Ah you right, good point!

torchsde/_core/adjoint.py
patrick-kidger: Fair enough.

torchsde/_core/base_sde.py
patrick-kidger: There's still an unused prior_drift argument.

torchsde/_core/methods/srk.py
patrick-kidger: Makes sense. We already try to factor as much of the noise type abstraction into the SDE (and out of the solvers) as it is, after all.

patrick-kidger:
Just spotted, both `sdeint` and its adjoint have the docstring for `names` still specifying the possibility of prior drift.

torchsde/_core/base_sde.py
lxuechen: Done.

lxuechen:
> Just spotted, both `sdeint` and its adjoint have the docstring for `names` still specifying the possibility of prior drift.

Done.

patrick-kidger:
This PR looks good to me. I've not hit the merge button as I assume the plan is to merge #40 first, so that the diagnostics can be run.

lxuechen:
Thanks for the reviews @patrick-kidger! All rates preserved after running the diagnostics. Merging now.

PR #38, opened 2020-08-24
Attempt to fix latent_sde

mtsokol:
Hi!
Details in: https://github.com/google-research/torchsde/pull/24#issuecomment-678782721

Current `dev` version of `latent_sde.py` is failing with:
```
...
logqp = [torch.stack(logqp_i, dim=0) for logqp_i in logqp]
RuntimeError: stack expects each tensor to be equal size, but got [512] at entry 0 and [] at entry 2
```
Tried to fix it by inspecting introduced changes. With this change it runs (checked first 3 generated plots). 
Is it OK? (If yes then I can refactor `integrate` also)

### DIFF
--- a/torchsde/_core/base_solver.py
+++ b/torchsde/_core/base_solver.py
@@ -211,8 +211,13 @@ class BaseSDESolver(metaclass=better_abc.ABCMeta):
                     prev_t, prev_y, prev_logqp = curr_t, curr_y, curr_logqp
                     curr_y, curr_logqp = self.step_logqp(curr_t, next_t, curr_y, curr_logqp)
                     curr_t = next_t
-            ret_y, ret_logqp = interp.linear_interp_logqp(t0=prev_t, y0=prev_y, logqp0=prev_logqp, t1=curr_t,
-                                                          y1=curr_y, logqp1=curr_logqp, t=out_t)
+        
+            if curr_t - out_t < 1e-7 or out_t - prev_t < dt_min:
+                ret_y, ret_logqp = interp.linear_interp_logqp(t0=prev_t, y0=prev_y, logqp0=prev_logqp, t1=curr_t,
+                                                              y1=curr_y, logqp1=curr_logqp, t=out_t)
+            else:
+                ret_y, ret_logqp = self.step_logqp(prev_t, out_t, prev_y, prev_logqp)
+
             ys.append(ret_y)
             [logqp_i.append(ret_logqp_i) for logqp_i, ret_logqp_i in zip(logqp, ret_logqp)]
 


patrick-kidger:
So the `logqp` aspects of the code base are proving to be a hassle to maintain, so (at least for now) we're deprecating them. (To be merged reasonably shortly in #37.)

Good news is, the bug you've observed shouldn't be a problem in the non-logqp case, at least.

This is actually why Chen set you might like to see if `latent_sde.py` can be replicated without using `logqp`. I think he was just suggesting manually writing out the logqp calculation in the SDE's drift, but FWIW I feel like we should actually be able to add do this by introducing a wrapper SDE (a bit like ForwardSDE, but exact form TBD).
The challenge is going to be doing so efficiently without repeated drift/diffusion evaluations, and exactly what that looks like I'm not sure. Any thoughts on that would be welcomed.

mtsokol:
Thanks for clarification! 
With fix I've experimented for a while and another SDE wrapper sounds cool - where extended drift will be evaluated each step but for `step` method we will be passing `g` and `f` without extension so dimensions are correct (so simply wrapping `g` and `f` won't work)
Leaving it for now.

lxuechen:
This can be achieved by simply augmenting the drift with the extra log-ratio term and the diffusion with a row of 0s. The noise type shouldn't change either. The Brownian motion doesn't even need to be modified.

mtsokol:
Just asking for learning purposes - so following instructions in https://github.com/google-research/torchsde/pull/24#issuecomment-678729732 would that refactor be something like: https://github.com/google-research/torchsde/compare/dev...mtsokol:dev-latent-experiment?expand=1 ?

So drift wrapper is augmenting result with `(f - h) / g` that is being computed only in forward stage. Due to this fact I had to create artificial column in y0 that is being ignored so input dimensions pass checks. Of course that's all hacky - asking only whether is it good direction?
Also should prior drift be augmented as posterior drift?

mtsokol:
Hm I've tried to do it two ways. First idea was as you described: I've augmented drift with `(f - h) / g` expression and diffusion with zeros. I haven't recreated original results yet. 
The first caveat for it is that for `sample_p` and `sample_q` the logqp part is also evaluated needlessly and second is that for current `master` version `srk` is used as a solver but for logqp computation it looks like `euler` hardcoded in `step_logqp`. Here in my implementation `srk` is used for both.
After 100 iterations I got sth like this (not great):


Second idea to avoid both problems is to perform separate `sdeint` in `forward` where drift and diffusion will be proper for logqp and `euler` method will be used, like:
```
zs = sdeint(...)
logqp = sdeint(..., method='euler', names={'drift': 'f_qp', 'diffusion': 'g_qp'}, ...)
```
I don't see other possibility without modifying internals like `base_solver`. Also both codes feel way slower per iteration than `master` version. What could I modify in terms of speed? E.g. use BrownianPath or different levy area approximation in BrownianInterval?

WDYT?

Also what I wanted to ask is how do you do such experiments? For me waiting e.g. 100 iterations on my computer to see some results takes eternity and checking few modifications takes few hours. Do you use e.g. some platform or GCP instance to run these faster?

lxuechen:
> First idea was as you described: I've augmented drift with (f - h) / g expression and diffusion with zeros. I haven't recreated original results yet.

This is not what I suggested. I explicitly mentioned that the drift could be augmented with the *gradient field* of the log-ratio term. The log-ratio term is 1/2 \int_0^T \norm{ (f - h) / g }^2 \dt, in which case the gradient field is 1/2 \norm{ (f - h) / g }^2. Integrating this vector field along with the original system with `sdeint` gives the correct integral. 

> The first caveat for it is that for sample_p and sample_q the logqp part is also evaluated needlessly 

This should not be a problem. E.g. you could do training with so constructed augmented functions `f_aug` and `g_aug`, and do the sampling with the original functions `f` and `g`. Obviously, this would require specifying the `names` argument, and that was precisely the rationale for including this as part of the API. Since the dimensions of the Brownian motion is different in two cases, you would need two of those. 

> second is that for current master version srk is used as a solver but for logqp computation it looks like euler hardcoded in step_logqp.

This also shouldn't matter much. Using a higher-order solver would only produce more accurate solutions. 

> After 100 iterations I got sth like this (not great):

This is likely due to the above mentioned bug in your code. Also, 100 iterations doesn't seem sufficient to me. My experience with the default hparams is that you would need around 300-400 to get some sensible data fitting. 

> Second idea to avoid both problems is to perform separate sdeint in forward where drift and diffusion will be proper for logqp and euler method will be used, like:

I don't like this approach; if you set up `f_aug` and `g_aug` properly, you shouldn't need this. 

> Also what I wanted to ask is how do you do such experiments? For me waiting e.g. 100 iterations on my computer to see some results takes eternity and checking few modifications takes few hours.

I'd suggest first try disabling the adjoint and using a low order solver. You could also increase the step size. When you find things working, switch back to the original hparams and train again. 

In general, if you find patching a fix too challenging, feel free to leave this off, and I can do this near the end of this month.

mtsokol:
Thank you for clarification! I'll give it one more try.

lxuechen:
About speeding up experiments, if you're running on a desktop CPU, it might also make sense to reduce the batch size, e.g. 128/256 should also work. You could also turn off adaptive time-stepping as there's currently an issue with it specifically for SRK. 

Hope this clarify things and feel free to post additional messages if you get blocked again. I'm on vacation this week, but I'll try to respond from time to time if possible.

Issue #15, opened 2020-08-14
Switch Brownians to interval interface


patrick-kidger:
Done:
- Brownian
- BrownianInterval
- BrownianPath
- BrownianTree
- calls in adjoint

Will do:
- calls in solvers

I know you're working on the tests so I've left those alone. Other calls that may be best to let you handle:
- calls in examples/demo.ipynb
- calls in tests/problems.py
- calls in tests/test_brownian_path.py
- calls in tests/test_brownian_tree.py
- calls in brownian_lib

patrick-kidger:
@lxuechen Benchmarks look rather encouraging for the BrownianInterval. e.g. see below. (They all exhibit the same behaviour.)

I'd note that the number of time steps is very low in the benchmarks (only 100).

![image](https://user-images.githubusercontent.com/33688385/90274146-fa3c5380-de57-11ea-9c99-5404226f145c.png)

lxuechen:
This is encouraging! Would you have an idea as to where the gain is coming from? 

I would guess it's coming from 1) using `__slots__`, and 2) explicitly modeling increments. 

How much different are the plots when these bms are plugged in the solvers?

patrick-kidger:
I can think of a couple other things for certain cases:
- For random access: not having to do list inserts, which are O(n) on average.
- For the GPU: not to having to do a `.to()` copy and instead initialising right on the GPU via `device=`. TBH this is really just a bug in the other implementations though.

The equivalent plots for the solvers look pretty much identical; I just picked one at random.

lxuechen:
Several side comments: 

> Switch to interval interface.

This breaks backwards compatibility. I'd be nice if we could have `tb` as an optional argument for now. 

> Removed 'trapezoidal_approx' as option from SRK solvers

This [line](
https://github.com/google-research/torchsde/blob/33fa1cdc37f018ea2122aa91853444720b1dbf99/torchsde/_core/methods/srk.py#L55) might be wrong if the second output of `bm` is H in James' thesis, since I_k0 isn't exactly the space-time Levy area. 

A side comment is that I don't totally support merging the SRK methods for now. But I think we should be able to get rid of the tableuas folder if you find it disturbing. 

Also the docstring needs to change for that file.

patrick-kidger:
> 
> 
> Several side comments:
> 
> > Switch to interval interface.
> 
> This breaks backwards compatibility. I'd be nice if we could have `tb` as an optional argument for BPath and BTree for now.

Done.
> 
[...]

PR #39, opened 2020-08-24
Euler-Heun method

mtsokol:
Hi!
Just euler-heun method.
Tried to run it but it hangs on BrownianInterval so related to #36

### DIFF
--- a/diagnostics/stratonovich_additive.py
+++ b/diagnostics/stratonovich_additive.py
@@ -28,6 +28,7 @@ from .utils import to_numpy, makedirs_if_not_found, compute_mse
 
 from tests.problems import Ex3Additive
 
+
 def inspect_sample():
     batch_size, d, m = 32, 1, 5
     steps = 10
@@ -47,35 +48,33 @@ def inspect_sample():
         ys_heun = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='heun')
         ys_euler_heun = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='euler_heun')
         ys_midpoint = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='midpoint')
-        ys_milstein_strat = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein')
-        ys_mil_strat_grad_free = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein',
+        ys_milstein = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein')
+        ys_milstein_grad_free = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein',
                                         options={'grad_free': True})
         ys_true = sdeint(sde_ito, y0=y0, ts=ts, dt=1e-3, bm=bm, method='euler')
-
-        print(ys_heun.shape, ys_heun.squeeze().shape)
 
         ys_heun = ys_heun.squeeze().t()
         ys_euler_heun = ys_euler_heun.squeeze().t()
         ys_midpoint = ys_midpoint.squeeze().t()
-        ys_milstein_strat = ys_milstein_strat.squeeze().t()
-        ys_mil_strat_grad_free = ys_mil_strat_grad_free.squeeze().t()
+        ys_milstein = ys_milstein.squeeze().t()
+        ys_milstein_grad_free = ys_milstein_grad_free.squeeze().t()
         ys_true = ys_true.squeeze().t()
 
-        ts_, ys_heun_, ys_euler_heun_, ys_midpoint_, ys_milstein_strat_, ys_mil_strat_grad_free_, ys_true_ = to_numpy(
-            ts, ys_heun, ys_euler_heun, ys_midpoint, ys_milstein_strat, ys_mil_strat_grad_free, ys_true)
+        ts_, ys_heun_, ys_euler_heun_, ys_midpoint_, ys_milstein_, ys_milstein_grad_free_, ys_true_ = to_numpy(
+            ts, ys_heun, ys_euler_heun, ys_midpoint, ys_milstein, ys_milstein_grad_free, ys_true)
 
     # Visualize sample path.
     img_dir = os.path.join('.', 'diagnostics', 'plots', 'stratonovich_additive')
     makedirs_if_not_found(img_dir)
 
-    for i, (ys_heun_i, ys_euler_heun_i, ys_midpoint_i, ys_milstein_strat_i, ys_mil_strat_grad_free_i, ys_true_i) in enumerate(
-            zip(ys_heun_, ys_euler_heun_, ys_midpoint_, ys_milstein_strat_, ys_mil_strat_grad_free_, ys_true_)):
+    for i, (ys_heun_i, ys_euler_heun_i, ys_midpoint_i, ys_milstein_i, ys_milstein_grad_free_i, ys_true_i) in enumerate(
+            zip(ys_heun_, ys_euler_heun_, ys_midpoint_, ys_milstein_, ys_milstein_grad_free_, ys_true_)):
         plt.figure()
         plt.plot(ts_, ys_heun_i, label='heun')
         plt.plot(ts_, ys_euler_heun_i, label='euler_heun')
         plt.plot(ts_, ys_midpoint_i, label='midpoint')
-        plt.plot(ts_, ys_milstein_strat_i, label='milstein_strat')
-        plt.plot(ts_, ys_mil_strat_grad_free_i, label='milstein_strat_grad_free')
+        plt.plot(ts_, ys_milstein_i, label='milstein_strat')
+        plt.plot(ts_, ys_milstein_grad_free_i, label='milstein_strat_grad_free')
         plt.plot(ts_, ys_true_i, label='true')
         plt.legend()
         plt.savefig(os.path.join(img_dir, f'{i}'))
@@ -92,8 +91,8 @@ def inspect_strong_order():
     heun_mses_ = []
     euler_heun_mses_ = []
     midpoint_mses_ = []
-    milstein_strat_mses_ = []
-    mil_strat_grad_free_mses_ = []
+    milstein_mses_ = []
+    milstein_grad_free_mses_ = []
 
     with torch.no_grad():
         bm = BrownianInterval(t0=ts[0], t1=ts[-1], shape=(batch_size, m), dtype=y0.dtype, device=device,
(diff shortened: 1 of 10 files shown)


torchsde/_core/methods/euler_heun.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+from .. import base_solver
+from ...settings import SDE_TYPES, NOISE_TYPES, LEVY_AREA_APPROXIMATIONS
+
+
+class EulerHeun(base_solver.BaseSDESolver):
+    weak_order = 1.0
+    sde_type = SDE_TYPES.stratonovich
+    noise_types = (NOISE_TYPES.additive, NOISE_TYPES.diagonal, NOISE_TYPES.general, NOISE_TYPES.scalar)
```
patrick-kidger, line 22: This can be changed to `NOISE_TYPES.all()`. (And indeed should probably be changed on some of the other solvers as well.)

mtsokol:
@patrick-kidger Hi! Just rebased - now it all works:

torchsde/_core/methods/euler_heun.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            self.strong_order = 1.0
+        super(EulerHeun, self).__init__(sde=sde, **kwargs)
+
+    def step(self, t0, t1, y0):
+        dt = t1 - t0
+        I_k = self.bm(t0, t1)
+
+        f_eval = self.sde.f(t0, y0)
+        g_prod_eval = self.sde.g_prod(t0, y0, I_k)
```
patrick-kidger, line 37: It looks like we're standardising on a convention of calling these local variables `f` and `g_prod`. (Similarly for `g_prod_eval_prime `.) c.f. Heun.

patrick-kidger (commented):
Other than the minor comments, LGTM. Thanks for implementing.
The fix for BrownianInterval is now in dev so you should be able to run the diagnostics. What do they look like?

mtsokol:
@patrick-kidger I've rewritten stratonovich diagnostics to scalar noise type (found one issue). Should I commit those files also? (it's basically copy paste of diagonal except for problem) Or only post plots here?
(and the same question regarding additive noise type)

patrick-kidger:
As you've got them, let's include the files as well. We'll probably be interested in running those later too.
(In fact the diagnostics all share a lot of common code that should be factored out somewhere, but that's not super important now.)

mtsokol:
Ok, I included stratonovich diagnostics for additive and scalar.
For both basic plots look OK.

Rates plots look like this:
- scalar

- additive


There's an issue with `milstein` solver - for `scalar` noise dimensions does not conform to compute `Y_prime` as `g` it's one additional dimension bigger (included fix but it's a quick one just to make it pass).

And looking at rates `milstein` for `additive` it's wrong but haven't figured it out yet. Btw both `grad-free` and default use `gdg_prod_default` - according to comment `grad_free` path isn't ready yet, right?

torchsde/_core/methods/euler_heun.py
mtsokol: Done!

torchsde/_core/methods/euler_heun.py
```diff
@@ -0,0 +1,45 @@
...
+
+class EulerHeun(base_solver.BaseSDESolver):
+    weak_order = 1.0
+    sde_type = SDE_TYPES.stratonovich
+    noise_types = NOISE_TYPES.all()
+    levy_area_approximations = LEVY_AREA_APPROXIMATIONS.all()
+
+    def __init__(self, sde, **kwargs):
+        if sde.noise_type == NOISE_TYPES.additive:
+            self.strong_order = 0.5
```
lxuechen, line 27: I think the strong order of E-H should be 1.0 for additive noise. It seems your plots also verify this.

torchsde/_core/methods/milstein.py
```diff
@@ -60,7 +60,7 @@ def step(self, t0, t1, y0):
         g_prod_I_k = self.sde.g_prod(t0, y0, I_k)
 
         if self.options[METHOD_OPTIONS.grad_free]:
-            g = self.sde.g(t0, y0)
+            g = self.sde.g(t0, y0) if self.sde.noise_type is not NOISE_TYPES.scalar else self.sde.g(t0, y0)[:,:,0]
```
lxuechen, line 63: Could you do something similar to [this](https://github.com/google-research/torchsde/blob/19062d5e7ce8a13b0e71092e2153c6f74a947943/torchsde/_core/methods/srk.py#L74)?

tests/problems.py
```diff
@@ -152,6 +159,15 @@ def analytical_sample(self, y0, ts, bm):
     def nfe(self):
         return self._nfe
 
+class Ex1Additive(Ex1):
```
lxuechen, line 162: I don't think you can make `Ex1` to have additive noise like this. Additive noise SDEs are defined such that the diffusion should not depend on the state, i.e. the derivative be 0. The original `Ex1` doesn't satisfy this.

lxuechen:
So for additive noise SDEs, the diffusion doesn't depend on the state. So the Ito interpretation is the same as the Strat interpretation, and the Milstein schemes should all have the same strong order as the Euler scheme.

torchsde/_core/methods/euler_heun.py
mtsokol: Sure, done.

torchsde/_core/methods/milstein.py
mtsokol: Done

tests/problems.py
mtsokol: Sure, done.

mtsokol:
Thanks for review - now I got it. Updated plots above - now it all looks right.

torchsde/_core/methods/euler_heun.py
```diff
@@ -0,0 +1,45 @@
...
+    weak_order = 1.0
+    sde_type = SDE_TYPES.stratonovich
+    noise_types = NOISE_TYPES.all()
+    levy_area_approximations = LEVY_AREA_APPROXIMATIONS.all()
+
+    def __init__(self, sde, **kwargs):
+        if sde.noise_type == NOISE_TYPES.additive:
+            self.strong_order = 1.0
+        else:
+            self.strong_order = 1.0
```
patrick-kidger, line 29: The above if statement is superfluous. Also, the strong order should be 0.5 for general noise.

mtsokol: Ah, sorry for that, fixed. 
Also I've started working on a benchmark from [19] from the paper ("Smoking adjoints"). For now I've tried to follow exactly their formulas and they use Euler solver in log form - I need to compute `g` there but for `adjoint_sdeint` it fails as it's not supported due to efficiency. What can I do about it? Can `g` somehow be computed for `adjoint_sdeint`? (I've also tried to make it as a field in sde and access it from `_base_sde` from solver but as in forward it's wrapped: `ForwardSDE(base_sde)` in backward I think it's `AdjointSDE(ForwardSDE(base_sde))`)

patrick-kidger:
This LGTM. @lxuechen happy for me to pull the squash+merge trigger?

torchsde/_core/methods/euler_heun.py
patrick-kidger: So back when that paper was written, "adjoint" meant "backprop through the solver".
These days, confusingly, we've appropriated the term to mean "don't backprop through the solver, and solve a differential equation instead".
As a result that paper is basically just about the now well-understood advantages of using backpropgation over forward-mode differentiation. (At least in this context.)

So they don't actually use the adjoint SDE at all, and what they do is equivalent to just using our `sdeint` in the normal way.

mtsokol: Yes, I got it while reading the paper. What I'm trying is to extend that benchmark from "smoking adjoints" with stochastic adjoint from here so I want to do it `forward` vs `sdeint` vs `adjoint_sdeint`. 
And due to the fact that in custom log_euler solver that I created I'm using `g` the described error occurs.

patrick-kidger: I'm not sure what your distinction between `forward` and `sdeint` is.

You shouldn't need to create a custom solver I don't think - IIRC in that paper they use an SDE to model log-rates; solve that with Euler--Maruyama; take an exponential if you want the rates themselves.

Likewise, you shouldn't need to use `g` in the adjoint: the standard EM scheme only needs `g_prod`.

Does that make sense / am I understanding you correctly? If not perhaps can you provide some references to where in the SA paper you're looking? (There's a couple copies of that paper out there, I tend to look at this one - https://www0.gsb.columbia.edu/faculty/pglasserman/Other/RiskJan2006.pdf)

mtsokol: Yeah I was reading exactly that one. 

By `forward` method I mean computing `dV/dS` with `(sdeint(S+h) - sdeint(S-h)) / 2h`, by `sdeint` ("adjoint" in "smoking adjoints", "backprop through the solver" here 😄) `dV/dS` is `sol = sdeint(); torch.grad(sol, S)` and last one `adjoint_sdeint` is `dV/dS` is `sol = adjoint_sdeint(); torch.grad(sol, S)`. (I hope I didn't mess it up). Also I've got PoC and they yield same gradient result.

Code is similar to this (as `forward` vs `sdeint`): https://github.com/mgroncki/AdjointMonteCarlo/blob/master/PlainVanillaOption/blackMC.cpp

Thanks for help! I will try it that way. To explain myself the log-Euler solver was used in paper and also in references in `Monte Carlo Methods In Financial Engineering` (page 398) directly for that model.

mtsokol: But I think I need a bit more time to fully figure all these methods out on paper. 
The chapter 7. of that book also makes comparison of two methods: the finite-difference approach defined as `(Y(S+h) - Y(S-h)) / 2h` and `pathwise` mehtod.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+import tqdm
+from scipy import stats
+
+from tests.basic_sde import AdditiveSDE
+from torchsde import sdeint, BrownianInterval
+from torchsde.settings import LEVY_AREA_APPROXIMATIONS
+from .utils import to_numpy, makedirs_if_not_found, compute_mse
+
+from tests.problems import Ex3Additive
+
```
lxuechen, line 30: Typical style guides enforce two blank lines between last import and first function/class definition.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+    with torch.no_grad():
+        bm = BrownianInterval(t0=ts[0], t1=ts[-1], shape=(batch_size, m), dtype=y0.dtype, device=device,
+                              levy_area_approximation=LEVY_AREA_APPROXIMATIONS.space_time)
+
+        ys_heun = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='heun')
+        ys_euler_heun = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='euler_heun')
+        ys_midpoint = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='midpoint')
+        ys_milstein_strat = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein')
+        ys_mil_strat_grad_free = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein',
```
lxuechen, line 51: Why not name this `ys_milstein_strat_grad_free`, i.e. use the full word of "milstein"? You could also probably just remove the `strat` part in the names, since this is assumed for all except `ys_true`.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+                                        options={'grad_free': True})
+        ys_true = sdeint(sde_ito, y0=y0, ts=ts, dt=1e-3, bm=bm, method='euler')
+
+        print(ys_heun.shape, ys_heun.squeeze().shape)
```
lxuechen, line 55: Please be sure to remove all print/debugging statements when submitting future PRs.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+
+        ys_heun = ys_heun.squeeze().t()
+        ys_euler_heun = ys_euler_heun.squeeze().t()
+        ys_midpoint = ys_midpoint.squeeze().t()
+        ys_milstein_strat = ys_milstein_strat.squeeze().t()
+        ys_mil_strat_grad_free = ys_mil_strat_grad_free.squeeze().t()
```
lxuechen, line 61: Same naming problem.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+        ys_true = ys_true.squeeze().t()
+
+        ts_, ys_heun_, ys_euler_heun_, ys_midpoint_, ys_milstein_strat_, ys_mil_strat_grad_free_, ys_true_ = to_numpy(
```
lxuechen, line 64: This line is too long. Unlike google style, I use a 120 character line wrap, and I think Patrick does something similar. I'd recommend stick with this or use 80/100. Anything mostly works so long as it's under 120.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+            ts, ys_heun, ys_euler_heun, ys_midpoint, ys_milstein_strat, ys_mil_strat_grad_free, ys_true)
+
+    # Visualize sample path.
+    img_dir = os.path.join('.', 'diagnostics', 'plots', 'stratonovich_additive')
+    makedirs_if_not_found(img_dir)
+
+    for i, (ys_heun_i, ys_euler_heun_i, ys_midpoint_i, ys_milstein_strat_i, ys_mil_strat_grad_free_i, ys_true_i) in enumerate(
```
lxuechen, line 71: Line too long.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+        bm = BrownianInterval(t0=ts[0], t1=ts[-1], shape=(batch_size, m), dtype=y0.dtype, device=device,
+                              levy_area_approximation=LEVY_AREA_APPROXIMATIONS.space_time)
+
+        for dt in tqdm.tqdm(dts):
+            # Only take end value.
+            _, ys_heun = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='heun')
+            _, ys_euler_heun = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='euler_heun')
+            _, ys_midpoint = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='midpoint')
+            _, ys_milstein_strat = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein')
+            _, ys_mil_strat_grad_free = sdeint(sde, y0=y0, ts=ts, dt=dt, bm=bm, method='milstein', options={'grad_free': True})
```
lxuechen, line 108: Same naming problem.

diagnostics/stratonovich_additive.py
```diff
@@ -0,0 +1,166 @@
...
+            _, ys_analytical = sde.analytical_sample(y0=y0, ts=ts, bm=bm)
+
+            heun_mse = compute_mse(ys_heun, ys_analytical)
+            euler_heun_mse = compute_mse(ys_euler_heun, ys_analytical)
+            midpoint_mse = compute_mse(ys_midpoint, ys_analytical)
+            milstein_strat_mse = compute_mse(ys_milstein_strat, ys_analytical)
+            mil_strat_grad_free_mse = compute_mse(ys_mil_strat_grad_free, ys_analytical)
```
lxuechen, line 115: Same naming problem.

diagnostics/stratonovich_additive.py
lxuechen, line 108: Line too long.

lxuechen:
The code in general is okay. I'd recommend using a linter with the character limits and blank line between imports/functions enforced. 

Thanks again for the putting in the effort in implementing these methods.

diagnostics/stratonovich_additive.py
mtsokol: Done!

diagnostics/stratonovich_additive.py
mtsokol: Done!

diagnostics/stratonovich_additive.py
mtsokol: Yikes, removed.

diagnostics/stratonovich_additive.py
mtsokol: Done!

diagnostics/stratonovich_additive.py
mtsokol: Sure - will stick to 120 then! This line is 112 - should I wrap it anyway?

diagnostics/stratonovich_additive.py
mtsokol: This line is 120 - should I wrap it anyway?

diagnostics/stratonovich_additive.py
mtsokol: Done!

diagnostics/stratonovich_additive.py
mtsokol: Done!

diagnostics/stratonovich_additive.py
mtsokol: Sure, done!

mtsokol:
Thanks for review! My main PL is Scala and python integration in my VS code barely works but I will setup that linter options to avoid those comments in the future.

mtsokol:
Just one moment - `milstein` naming should be fixed also in `diagonal` file.

mtsokol:
@lxuechen OK, now it's ready.

mtsokol:
P.S. (asking out of curiosity) Do you have a roadmap for merging `dev` into `master` and cutting new release?

patrick-kidger:
@mtsokol 
Loosely speaking yes, although @lxuechen and I haven't fixed things in detail.

I think we'd like to tick off the remaining things in #15. Fortunately, I think we've now done all the complicated ones, and it's mostly just grunt work remaining.
My hope is for a new release this month. (I'm only speaking personally there; Chen and I haven't discussed release schedules yet.)

lxuechen:
@mtsokol @patrick-kidger 

I agree mostly with Patrick. Aside from the technical problems in #15, on a high level, I think the most crucial missing things are
- documentation (docstring, README, additional doc site generated from docstring?)
- tests, tests, and tests
- benchmarks
- continuous integration
- backward comp fixes

I think we should be able to pull off another release in this month, if things go well. Though, I also think we shouldn't rush things by too much and only merge when we're genuinely ready.

PR #40, opened 2020-08-24
Improved heuristic for BrownianInterval's dependency tree.

### LINKED ISSUE #36: [On dev branch] New BrownianInterval hangs when `dt` is very small

patrick-kidger:
Fixes #36.
What was meant to be a small speedup ended up being a Russian roulette between "small speedup" and "huge slowdown".

### DIFF
--- a/torchsde/_brownian/brownian_interval.py
+++ b/torchsde/_brownian/brownian_interval.py
@@ -109,6 +109,7 @@ class _Interval:
         return self._top.increment_and_space_time_levy_area_cache(self)
 
     def increment_and_space_time_levy_area_uncached(self):
+        # TODO: switch this over to a trampoline?
         if self._is_left:
             return self._parent.left_increment_and_space_time_levy_area()
         else:
@@ -408,7 +409,7 @@ class BrownianInterval(_Interval, base_brownian.BaseBrownian):
             raise ValueError(f"`levy_area_approximation` must be one of {LEVY_AREA_APPROXIMATIONS}, but got "
                              f"'{levy_area_approximation}'.")
 
-        self._dt = None
+        self._dt = dt
         self._cache_size = cache_size
         self.levy_area_approximation = levy_area_approximation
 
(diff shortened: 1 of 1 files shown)


lxuechen:
Could you explain a bit more what's changed? 

Also, this doesn't seem like a complete fix to me. It is still possible to get very small step sizes, and the issue will likely bug the updated BTree and BPath in the future.

patrick-kidger:
Sure.
So BInterval relies on a dependency structure between queried intervals as you know. If that's the only structure that is imposed, then a problem arises in that the backward pass ends up taking O(n^2) work in the length n, because you have to recompute your way forwards from the start to get to the queried interval, then step back a bit and repeat the whole procedure. (The cacheing offsets this somewhat but doesn't change the asymptotics.)
The resolution to this is to prespecify a tree-like dependency structure. This increases the cost of the forward pass from O(n) to O(nlogn), but decreases the cost of the backward pass to the same.
The problem _now_ is that we have to pick a resolution to create this dependency structure down to.
The optimal resolution is a function of the average step size in the solver, and the size of the cache.

The bit that's been changed is the heuristic for figuring out the average step size in the solver. Previously, it was very crude: if any given step was less than 0.1 the smallest step so far recorded, then a dependency tree was created down to that resolution.
But because the terminal step in the new solver could be arbitrarily small, then this final step would trigger such a (completely unnecessary) dependency tree creation.

The fix is to actually calculate the average interval length that it's been queried width, and only trigger a dependency tree creation is _that_ gets small enough. Just one tiny step isn't enough to trigger it.

This isn't an issue that should affect BTree or BPath - it's just an undesired interaction between the updated solver and an overly naive heuristic in BInterval.

lxuechen:
Ok, thanks for explaining the average step size computation part. LGTM.

I think at some point, I want to revisit the weird behavior that `dt` is very small at the end. This actually does affect the new BPath I wrote, since it currently relies on Cholesky, and it's proving to not be very numerically stable. Though, alternatively, I could also use the H generation formulae in James' thesis and get the same thing.

PR #42, opened 2020-08-25
Binterval recursion fix

patrick-kidger:
Rather than trampolining everything, I've taken the more straightforward approach of inlining a lot of the function indirections that were previously present. This should reduce the size of the call stack that's built up by a factor of something like six.

The code is largely cleaner for this change, as well, which is nice.

It does introduce one additional dependency: [boltons](https://boltons.readthedocs.io/en/latest/), which is a popular stdlib-like library for "things that are missing from the stdlib"; in this case a better LRU cache implementation. Obviously additional dependencies are to be avoided, but I think this one is reasonable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setuptools.setup(
                        optional=True)
     ],
     cmdclass={'build_ext': cpp_extension.BuildExtension},
-    install_requires=['torch>=1.6.0', 'blist', 'numpy>=1.19.1', 'boltons'],
+    install_requires=['torch>=1.6.0', 'blist', 'numpy>=1.19.1', 'boltons>=20.2.1'],
     python_requires='~=3.6',
     classifiers=[
         "Programming Language :: Python :: 3",
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setuptools.setup(
                        optional=True)
     ],
     cmdclass={'build_ext': cpp_extension.BuildExtension},
-    install_requires=['torch>=1.6.0', 'blist', 'numpy>=1.19.1'],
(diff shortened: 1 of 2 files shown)


torchsde/_brownian/brownian_interval.py
```diff
@@ -106,80 +101,67 @@ def increment_and_levy_area(self):
         return W, H, A
 
     def _increment_and_space_time_levy_area(self):
-        return self._top.increment_and_space_time_levy_area_cache(self)
-
-    def increment_and_space_time_levy_area_uncached(self):
         # TODO: switch this over to a trampoline?
```
lxuechen, line 104: If the plan is to not go for the trampoline approach, should this TODO be removed?

setup.py
```diff
@@ -67,7 +67,7 @@
                        optional=True)
     ],
     cmdclass={'build_ext': cpp_extension.BuildExtension},
-    install_requires=['torch>=1.6.0', 'blist', 'numpy>=1.19.1'],
+    install_requires=['torch>=1.6.0', 'blist', 'numpy>=1.19.1', 'boltons'],
```
lxuechen, line 70: Is there a particular min version for boltons that is required? This is mostly reflecting from our numpy version mismatch discussion a while ago.

torchsde/_brownian/brownian_interval.py
patrick-kidger: I'm leaving it there for now as I think adding a trampoline may still be useful later on, if any use case gives even deeper call stacks.

setup.py
patrick-kidger: Not that I'm aware of / have tried tracking down.

PR #43, opened 2020-08-31
Add log-ODE scheme and simplify typing.

lxuechen:
Main things here:

- Add log-ODE scheme obtained with Lie-Trotter splitting and midpoint as described in your draft
- Unify all the diagnostics files
- Minor tweaks for problems and utility functions in `diagnostics`
- Put all type info in `types.py` so that we don't have to import two modules elsewhere

Here's the strong order plot
rate

Off the top of my head, I'm not entirely sure what strong order to expect. We could also do some W_2 error analysis, and I've done this in the past with `pot` for my sampling paper, but it's rather slow to run, given that these are cubic time algorithm IIRC. On another note, Sinkhorn would produce extra error. 

Thinking out loud, possible sources of error just for this set of diagnostics are
- true solution approximation by midpoint
- not enough samples to approximate the mse


diagnostics/utils.py
```diff
@@ -13,36 +13,111 @@
...
-    assert mse.size() == ()
-    return mse
+def mse(x: Tensor, y: Tensor, norm_dim: Optional[int] = 1, mean_dim: Optional[int] = 0) -> np.ndarray:
+    """Compute mean squared error."""
+    return _to_numpy_single((torch.norm(x - y, dim=norm_dim) ** 2).mean(dim=mean_dim))
+
+
+def makedirs(*dirs: str):
+    """Create directories if not already present according to a sequence of strings."""
+    [os.makedirs(d) for d in dirs if not os.path.exists(d)]
```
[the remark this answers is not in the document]
patrick-kidger: Actually also, just checked - this is only called once and so can be inlined.

tests/problems.py
```diff
@@ -12,35 +12,44 @@
...
 
     def f(self, t, y):
-        del t
         self._nfe += 1
         return self.mu * y
 
     def f_corr(self, t, y):
-        del t
         self._nfe += 1
         return self.mu * y - .5 * (self.sigma ** 2) * y
```
patrick-kidger, line 50: None of the `f_corr` are used. I don't think the Strat diagnostics will be correct.
(Easiest thing would be to just assign the correct f in `__init__`)

diagnostics/inspect.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2020 Google LLC
```
patrick-kidger, line 1: Let's not pick module names overlapping with the stdlib.

torchsde/_core/methods/log_ode.py
```diff
@@ -0,0 +1,53 @@
...
+The scheme uses Levy area approximations.
+"""
+
+from .. import adjoint_sde
+from .. import base_solver
+from ...settings import SDE_TYPES, NOISE_TYPES, LEVY_AREA_APPROXIMATIONS
+
+
+class LogODEMidpoint(base_solver.BaseSDESolver):
+    strong_order = 0.5
```
patrick-kidger, line 26: This should get strong order 1.0 for every noise, I think.

torchsde/_core/methods/log_ode.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, sde, **kwargs):
+        if isinstance(sde, adjoint_sde.AdjointSDE):
+            raise ValueError(f"Log-ODE schemes cannot be used for adjoint SDEs, because they require "
+                             f"direct access to the diffusion, whilst adjoint SDEs rely on a more efficient "
+                             f"diffusion-vector product. Use a different method instead.")
+        super(LogODEMidpoint, self).__init__(sde=sde, **kwargs)
+
+    def step(self, t0, t1, y0):
+        dt = t1 - t0
+        W, _, A = self.bm(t0, t1, return_U=True, return_A=True)
```
patrick-kidger, line 41: This can just be `W, A = self.bm(t0, t1, return_A=True)`

torchsde/_core/methods/log_ode.py
```diff
@@ -0,0 +1,53 @@
...
+
+        y_prime = y0 + .5 * (self.sde.f(t0, y0) * dt + self.sde.g_prod(t0, y0, W))
+        t_prime = .5 * (t0 + t1)
```
patrick-kidger, line 44: Nitpick: I'd prefer to represent this calculation in the same way as midpoint, just for consistency's sake.

torchsde/settings.py
```diff
@@ -23,16 +23,16 @@ def __str__(cls):
...
 #  like this. (e.g. instead set METHODS = {'euler': Euler, ...} in methods/__init__.py)
-
-
 class METHODS(metaclass=ContainerMeta):
     euler = 'euler'
     milstein = 'milstein'
     srk = 'srk'
     midpoint = 'midpoint'
     heun = 'heun'
+    log_ode_midpoint = 'log_ode_midpoint'
```
patrick-kidger, line 35: The difference between midpoint and Heun here is essentially irrelevant. I'd like to add just `log_ode` as an alias for this. (And we can keep around `log_ode_midpoint` for the sake of some future `log_ode_heun`, if people really want that for some reason.)

patrick-kidger (commented):
When trying to check the diagnostics, I found I couldn't run them (or at least the Ito/Strat diagonal one) due to a RecursionError in BInterval, which didn't occur before this change.

I've spotted that our `types.py` module shares a name with one from the stdlib. (Whilst copying functionality from `typing`!) It should probably be renamed.

I'd say that the comment in `base_sde.py`: "# Assumes `a` is anti-symmetric and `_base_sde` is not of diagonal noise." isn't quite correct. It's the log-ODE method that is assuming that `a` is antisymmetric; in making that assumption it is valid to call something named `dg_ga`. Also I think it's okay to say that this assumes general noise, as all the other noise types get zero for this term.

patrick-kidger:
The 0.75 order is a bit mysterious looking. Adding another SDE to solve is likely to help diagnose if there's actually an issue - I recall that we got funny-looking convergence rates for one of the previous Strat solvers just because of the choice of SDE.

diagnostics/utils.py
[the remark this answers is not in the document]
lxuechen: Done.

diagnostics/utils.py
```diff
@@ -13,36 +13,111 @@
...
+
+
+def log(x: Union[Sequence[float], np.ndarray]) -> np.ndarray:
+    """Compute element-wise log of a sequence of floats."""
+    if not isinstance(x, np.ndarray):
+        return np.log(np.array(x))
+    return np.log(x)
+
+
+def half_log(x: Union[Sequence[float], np.ndarray]) -> np.ndarray:
```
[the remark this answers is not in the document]
lxuechen: Done removing.

diagnostics/utils.py
```diff
@@ -13,36 +13,111 @@
...
+def makedirs(*dirs: str):
+    """Create directories if not already present according to a sequence of strings."""
+    [os.makedirs(d) for d in dirs if not os.path.exists(d)]
+
+
+def log(x: Union[Sequence[float], np.ndarray]) -> np.ndarray:
+    """Compute element-wise log of a sequence of floats."""
+    if not isinstance(x, np.ndarray):
+        return np.log(np.array(x))
+    return np.log(x)
```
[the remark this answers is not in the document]
lxuechen: Done.

diagnostics/utils.py
```diff
@@ -13,36 +13,111 @@
...
+
+def half_log(x: Union[Sequence[float], np.ndarray]) -> np.ndarray:
+    """Compute element-wise half-log of a sequence of floats."""
+    return .5 * log(x)
+
+
+def linregress_slope(x, y):
+    """Return the slope of a least-squares regression for two sets of measurements."""
+    k, _, _, _, _ = stats.linregress(x, y)
+    return k
```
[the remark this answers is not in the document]
lxuechen: I simplified this, but I prefer to have this helper function to let the reader know this is the slope. This reads better than just `stats.linregress(x, y)[0]`, which doesn't imply much without reading the doc of `stats.linregress`.

tests/problems.py
lxuechen: There should not be any serious problems with the diagnostics, since I'm running Euler (for Ito) and Midpoint (for Strat) with very small step size for the ground-truth. This makes the code consistent.

lxuechen: Another note is that we should keep `f_corr` just in case we need them in the future.

torchsde/_core/methods/log_ode.py
lxuechen: I'm not too sure about that. Is Foster accurate enough?

torchsde/_core/methods/log_ode.py
lxuechen: Done.

torchsde/_core/methods/log_ode.py
lxuechen: Done.

lxuechen:
> When trying to check the diagnostics, I found I couldn't run them (or at least the Ito/Strat diagonal one) due to a RecursionError in BInterval, which didn't occur before this change.

Is the error a stackoverflow? I'm using Euler with very small step size for simulating the ground-truth now. 

> I've spotted that our types.py module shares a name with one from the stdlib.

I don't see this as a problem in this particular setting. Importing from this file would probably follow two styles:
- `import torchsde.types`
- `from torchsde.types import Tuple...`
The latter style is what we do in all files at the moment. I named this file as `types.py` originally following PyTorch.

torchsde/settings.py
lxuechen: Done.

diagnostics/inspect.py
lxuechen: Done.

diagnostics/stratonovich_scalar.py
```diff
@@ -12,153 +12,46 @@
...
+
+def main():
+    small_batch_size, large_batch_size, d, m = 16, 8192, 3, 5
+    t0, t1, steps, dt = 0., 2., 10, 1e-1
+    ts = torch.linspace(t0, t1, steps=steps, device=device)
+    dts = tuple(2 ** -i for i in range(1, 8))  # For checking strong order.
+    sde = Ex1Scalar(d=d, sde_type=SDE_TYPES.stratonovich).to(device)
+    # TODO: Add gradient-free Milstein.
+    #  Either split out g-f Milstein to be a separate method or change how inspection functions work.
+    methods = ('heun', 'euler_heun', 'midpoint', 'milstein')
```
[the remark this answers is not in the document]
lxuechen: Done.

diagnostics/stratonovich_scalar.py
```diff
@@ -12,153 +12,46 @@
...
-    plt.legend()
-
-    img_dir = os.path.join('.', 'diagnostics', 'plots', 'stratonovich_scalar')
-    makedirs_if_not_found(img_dir)
-    plt.savefig(os.path.join(img_dir, 'rate'))
-    plt.close()
+    y0 = torch.full((large_batch_size, d), fill_value=0.1, device=device)
+    bm = BrownianInterval(
+        t0=t0, t1=t1, shape=(large_batch_size, 1), dtype=y0.dtype, device=device,
+        levy_area_approximation=LEVY_AREA_APPROXIMATIONS.space_time, pool_size=16
```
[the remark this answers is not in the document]
lxuechen: Done.

lxuechen:
> The 0.75 order is a bit mysterious looking. Adding another SDE to solve is likely to help diagnose if there's actually an issue - I recall that we got funny-looking convergence rates for one of the previous Strat solvers just because of the choice of SDE.

I agree. I'll also check the rate for a simple SDE where I can check the Lipschitz constants of the coefficients and their derivatives.

patrick-kidger:
So - I think I agree that log-ODE w/ Levy area approximation could get sub-1.0 convergence.
(I'd note that what's there still needs tweaking to be 0.5 for general noise and 1.0 otherwise, where it just reduces to midpoint anyway.)

Regarding the rates - I looked at the SDE that this is tested on and I think I see a `y^2` (cross-)term. The product operation `x, y -> xy` isn't Lipschitz as a map `R^2 -> R`. I think if we have a non-Lipschitz vector field then all bets are off, and we should probably use something else for the rate diagnostics anyway?

diagnostics/stratonovich_general.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def main():
+    small_batch_size, large_batch_size, d, m = 16, 8192, 3, 5
+    t0, t1, steps, dt = 0., 2., 10, 1e-1
+    ts = torch.linspace(t0, t1, steps=steps, device=device)
+    dts = tuple(2 ** -i for i in range(1, 8))  # For checking strong order.
+    sde = Ex4(d=d, m=m, sde_type=SDE_TYPES.stratonovich).to(device)
+    methods = ('midpoint', 'log_ode')
+    img_dir = os.path.join('.', 'diagnostics', 'plots', 'stratonovich_general')
```
patrick-kidger, line 33: This depends upon the run location. Can this be switched to e.g:
```
import pathlib
here = pathlib.Path(__file__).resolve().parent
img_dir = here / 'plots/ ...
```

lxuechen: Good point. Fixed with 
```
img_dir = os.path.join(os.path.dirname(__file__), 'plots', ...)
```

lxuechen:
> (I'd note that what's there still needs tweaking to be 0.5 for general noise and 1.0 otherwise, where it just reduces to midpoint anyway.)

I just noticed that the weak order is 2.0 with foster. Do you know the weak order with davie? 

> The product operation x, y -> xy isn't Lipschitz as a map R^2 -> R. I think if we have a non-Lipschitz vector field then all bets are off, and we should probably use something else for the rate diagnostics anyway?

Good point. Without taming, I agree this is a problem theoretically. I will have this fixed and also test the weak order.

patrick-kidger:
Hummm, I've not thought about weak order at all tbh. Where are you getting that it has weak order 2.0 from?

lxuechen:
> Where are you getting that it has weak order 2.0 from?

This is handwavy. But the intuition I have is that given how similar the formulae are compared to Talay, and that Foster's also matches the first two moments of Levy area, it seems unlikely that it should have a lower order. The draft you sent me a while ago has some mildly relevant comments in the Talay connections section.

patrick-kidger:
log-ODE is basically Milstein, but with just the antisymmetric part of the second order signature (=Levy area) as the correction, rather than the full second order signature. So my expectation is order 1.0 a la Milstein.
I'd also note that Talay is an Ito scheme, whilst log-ODE is Strat. IIRC most high-order weak schemes are using something clever about the martingale properties of Ito to do well. (Although I've never really looked into that.)

lxuechen:
This sounds reasonable to me. I'll change back the weak order to 1.0 for now.

lxuechen:
@mtsokol Letting you know there are quite some changes here related to the `diagnostics` folder.

PR #45, opened 2020-09-03
Bugfixes for log-ODE

patrick-kidger:
Note that I'm looking to merge this into `dev-log-ode`.

- I noted that the diagnostics didn't run in #43, due to a RecursionError in BInterval. The reason is that the access patterns are a bad fit for what BInterval is optimised for. I've fixed this here by reordering the calls in `inspect_strong_order`.*
- I've also increased the smallest step size in the diagnostics from 2**-8 to 2**-7, as I started to observe the rates misbehaving for small values, presumably because they were approaching the "analytical" discretisation.
- I've also fixed a bug introduced in #43 that prevented `test_strat.py` from running. (The SDE was set to Ito.)
- Fixed a bug that prevented log-ODE from using the default-created BrownianInterval because it didn't have the right `levy_area_approximation` set.

*The RecursionErrors here are a good indication that trampolining / tail recursion is probably worth adding to BInterval. I'll probably focus on other things(=running experiments for the paper) in the short term, but I'll aim to get that added in time for our next release.

### DIFF
--- a/diagnostics/inspection.py
+++ b/diagnostics/inspection.py
@@ -85,7 +85,7 @@ def inspect_orders(y0: Tensor,
 
     sde = copy.deepcopy(sde).requires_grad_(False)
     ts = torch.tensor([t0, t1], device=y0.device)
-    
+
     solns = [
         [sdeint(sde, y0, ts, bm, method=method, dt=dt, options=options_)[-1]
          for method, options_ in zip(methods, options)]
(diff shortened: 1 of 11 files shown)


lxuechen:
Thanks for the PR! Though, I actually have this access order fixed in my PR locally (not pushed). Would it make sense for me to push this fix in #43 after merging #44?

patrick-kidger:
Whatever you like really!

PR #48, opened 2020-09-04
Latent experiment

mtsokol:
Hi!

Following next instructions in https://github.com/google-research/torchsde/pull/38#issuecomment-686559686 here's my idea for that - I've looked what flow exactly is for previous version available on `master` and hopefully recreated it. 

That's a result after 300 iterations (still not similar to 300 in previous version):

WDYT?

btw. running it locally was still slow and burning laptop so I ended up running it on CGP instance (e2-highcpu-8) as I found they have student packs and got one iteration at 4sec.

### DIFF
--- a/examples/latent_sde.py
+++ b/examples/latent_sde.py
@@ -28,6 +28,7 @@ from torch.distributions import Normal, Laplace, kl_divergence
 
 from examples import utils
 from torchsde import sdeint, sdeint_adjoint, SDEIto, BrownianInterval
+from torchsde._core.misc import _stable_division
 
 Data = namedtuple('Data', ['ts_', 'ts_ext_', 'ts_vis_', 'ts', 'ts_ext', 'ts_vis', 'ys', 'ys_'])
 
@@ -74,15 +75,9 @@ class LatentSDE(SDEIto):
     def f_aug(self, t, y):
         y = y[:, 0:1]
         f, g, h = self.f(t, y), self.g(t, y), self.h(t, y)
-        z = f - h
-        g = torch.where(
-            g.abs() > 1e-7,
-            g,
-            torch.ones_like(g).fill_(1e-7) * g.sign()
-        )
-        z = z / g
-        z = .5 * (torch.norm(z, dim=1, keepdim=True) ** 2.)
-        return torch.cat([f, z], dim=1)
+        u = _stable_division(f - h, g)
+        u = .5 * torch.norm(u, dim=1, keepdim=True) ** 2.
+        return torch.cat([f, u], dim=1)
 
     def g(self, t, y):  # Shared diffusion.
         return self.sigma.repeat(y.size(0), 1)
@@ -96,7 +91,7 @@ class LatentSDE(SDEIto):
     def forward(self, ts, batch_size, eps=None):
         eps = torch.randn(batch_size, 1).to(self.qy0_std) if eps is None else eps
         y0 = self.qy0_mean + eps * self.qy0_std
-        y0 = torch.cat([y0, torch.zeros(batch_size, 1).to(self.qy0_std)], dim=1)
+        y0 = torch.cat([y0, torch.zeros(batch_size, 1).to(y0)], dim=1)
 
         qy0 = Normal(loc=self.qy0_mean, scale=self.qy0_std)
         py0 = Normal(loc=self.py0_mean, scale=self.py0_std)
--- a/examples/latent_sde.py
+++ b/examples/latent_sde.py
@@ -28,7 +28,6 @@ from torch.distributions import Normal, Laplace, kl_divergence
 
 from examples import utils
 from torchsde import sdeint, sdeint_adjoint, SDEIto, BrownianInterval
-from torchsde._core.misc import _stable_division
 
 Data = namedtuple('Data', ['ts_', 'ts_ext_', 'ts_vis_', 'ts', 'ts_ext', 'ts_vis', 'ys', 'ys_'])
 
@@ -75,7 +74,7 @@ class LatentSDE(SDEIto):
     def f_aug(self, t, y):
         y = y[:, 0:1]
         f, g, h = self.f(t, y), self.g(t, y), self.h(t, y)
-        u = _stable_division(f - h, g)
+        u = self._stable_division(f - h, g)
         u = .5 * torch.norm(u, dim=1, keepdim=True) ** 2.
         return torch.cat([f, u], dim=1)
 
@@ -87,6 +86,15 @@ class LatentSDE(SDEIto):
         g = self.g(t, y)
         z = torch.zeros_like(y)
         return torch.cat([g, z], dim=1)
+
+    @staticmethod
+    def _stable_division(x, y, epsilon=1e-7):
+        y = torch.where(
+            y.abs() > epsilon,
+            y,
+            torch.full_like(y, epsilon) * y.sign()
+        )
+        return x / y
 
     def forward(self, ts, batch_size, eps=None):
         eps = torch.randn(batch_size, 1).to(self.qy0_std) if eps is None else eps
--- a/examples/latent_sde.py
+++ b/examples/latent_sde.py
@@ -94,9 +94,9 @@ class LatentSDE(SDEIto):
         return torch.cat([g, z], dim=1)
 
     def forward(self, ts, batch_size, eps=None):
-        eps = torch.cat([torch.randn(batch_size, 1).to(self.qy0_std), torch.zeros(batch_size, 1)
-                        .to(self.qy0_std)], dim=1) if eps is None else eps
+        eps = torch.randn(batch_size, 1).to(self.qy0_std) if eps is None else eps
         y0 = self.qy0_mean + eps * self.qy0_std
+        y0 = torch.cat([y0, torch.zeros(batch_size, 1).to(self.qy0_std)], dim=1)
 
         qy0 = Normal(loc=self.qy0_mean, scale=self.qy0_std)
         py0 = Normal(loc=self.py0_mean, scale=self.py0_std)
@@ -109,9 +109,9 @@ class LatentSDE(SDEIto):
             zs = sdeint(self, y0, ts, method=args.method, dt=args.dt, adaptive=args.adaptive,
                                rtol=args.rtol, atol=args.atol, names={'drift': 'f_aug', 'diffusion': 'g_aug'})
         logqp = zs[-1, :, 1]
-        zs = zs[:, :, 0][:, :, None]
+        zs = zs[:, :, 0:1]
 
-        logqp = logqp.sum(0).mean(0)
+        logqp = logqp.mean(0)
         log_ratio = logqp0 + logqp  # KL(time=0) + KL(path).
 
         return zs, log_ratio
--- a/examples/latent_sde.py
+++ b/examples/latent_sde.py
@@ -71,11 +71,32 @@ class LatentSDE(SDEIto):
         inp = torch.cat((torch.sin(t), torch.cos(t), y), dim=-1)
         return self.net(inp)
 
+    def f_aug(self, t, y):
+        y = y[:, 0][:, None]
+        res = self.f(t, y)
+        z = res - self.h(t, y)
+        g = self.g(t, y)
+        g = torch.where(
+            g.abs() > 1e-7,
+            g,
+            torch.ones_like(g).fill_(1e-7) * g.sign()
+        )
+        z = z / g
+        z = .5 * (torch.norm(z, dim=1) ** 2.)[:, None]
+        return torch.cat([res, z], dim=1)
+
     def g(self, t, y):  # Shared diffusion.
         return self.sigma.repeat(y.size(0), 1)
 
+    def g_aug(self, t, y):
+        y = y[:, 0][:, None]
+        res = self.g(t, y)
+        z = torch.zeros_like(y)
+        return torch.cat([res, z], dim=1)
+
     def forward(self, ts, batch_size, eps=None):
-        eps = torch.randn(batch_size, 1).to(self.qy0_std) if eps is None else eps
+        eps = torch.cat([torch.randn(batch_size, 1).to(self.qy0_std), torch.zeros(batch_size, 1)
+                        .to(self.qy0_std)], dim=1) if eps is None else eps
         y0 = self.qy0_mean + eps * self.qy0_std
 
         qy0 = Normal(loc=self.qy0_mean, scale=self.qy0_std)
@@ -83,11 +104,14 @@ class LatentSDE(SDEIto):
         logqp0 = kl_divergence(qy0, py0).sum(1).mean(0)  # KL(time=0).
 
         if args.adjoint:
-            zs, logqp = sdeint_adjoint(self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive,
-                                       rtol=args.rtol, atol=args.atol)
+            zs = sdeint_adjoint(self, y0, ts, method=args.method, dt=args.dt, adaptive=args.adaptive,
+                                       rtol=args.rtol, atol=args.atol, names={'drift': 'f_aug', 'diffusion': 'g_aug'})
         else:
-            zs, logqp = sdeint(self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive,
-                               rtol=args.rtol, atol=args.atol)
+            zs = sdeint(self, y0, ts, method=args.method, dt=args.dt, adaptive=args.adaptive,
+                               rtol=args.rtol, atol=args.atol, names={'drift': 'f_aug', 'diffusion': 'g_aug'})
+        logqp = zs[:, :, 1]
+        zs = zs[:, :, 0][:, :, None]
+
         logqp = logqp.sum(0).mean(0)
         log_ratio = logqp0 + logqp  # KL(time=0) + KL(path).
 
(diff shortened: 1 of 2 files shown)


examples/latent_sde.py
```diff
@@ -71,23 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
         inp = torch.cat((torch.sin(t), torch.cos(t), y), dim=-1)
         return self.net(inp)
 
+    def f_aug(self, t, y):
+        y = y[:, 0][:, None]
```
lxuechen, line 75: You should be able to simplify this with `y = y[:, 0:1]`.

examples/latent_sde.py
```diff
@@ -71,23 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
...
+        res = self.f(t, y)
```
lxuechen, line 76: Why not just call this `f`?

examples/latent_sde.py
```diff
@@ -71,23 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
...
+        z = res - self.h(t, y)
+        g = self.g(t, y)
```
lxuechen, line 78: Why not do something like `f, g, h = self.f(t, y), self.g(t, y), self.h(t, y)`, so that we're clear there's a bunch of function evals?

examples/latent_sde.py
```diff
@@ -71,23 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
...
+        g = torch.where(
+            g.abs() > 1e-7,
+            g,
+            torch.ones_like(g).fill_(1e-7) * g.sign()
+        )
+        z = z / g
+        z = .5 * (torch.norm(z, dim=1) ** 2.)[:, None]
```
lxuechen, line 85: Use the `keepdim` option of `torch.norm`, instead of explicit dimension unsqueezing.

examples/latent_sde.py
```diff
@@ -71,23 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
...
+        return torch.cat([res, z], dim=1)
+
     def g(self, t, y):  # Shared diffusion.
         return self.sigma.repeat(y.size(0), 1)
 
+    def g_aug(self, t, y):
+        y = y[:, 0][:, None]
```
lxuechen, line 92: Same applies to this method.

examples/latent_sde.py
```diff
@@ -71,23 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
...
+                                       rtol=args.rtol, atol=args.atol, names={'drift': 'f_aug', 'diffusion': 'g_aug'})
         else:
-            zs, logqp = sdeint(self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive,
-                               rtol=args.rtol, atol=args.atol)
+            zs = sdeint(self, y0, ts, method=args.method, dt=args.dt, adaptive=args.adaptive,
+                               rtol=args.rtol, atol=args.atol, names={'drift': 'f_aug', 'diffusion': 'g_aug'})
+        logqp = zs[:, :, 1]
+        zs = zs[:, :, 0][:, :, None]
+
         logqp = logqp.sum(0).mean(0)
```
lxuechen, line 115: Now we should only take the last entry of `logqp`.

lxuechen:
I think now we're seeing something on the right track. The main issue here is that you should just take `zs[-1, :, 1]` as logqp to make it consistent with what we had before, as opposed to summing over the first dimension, which would mean inflating the KL divergence.

mtsokol:
@lxuechen I've applied all comments and run it but it doesn't seem good after 400 iterations:


Regarding taking last entry of `logqp` - I've tracked existing implementation and in https://github.com/google-research/torchsde/blob/master/torchsde/_core/base_solver.py#L227 `logqp` is appended after each step and I don't see where only last entry is retrieved.

examples/latent_sde.py
mtsokol: Done!

examples/latent_sde.py
mtsokol: Sure, done!

examples/latent_sde.py
mtsokol: Done!

examples/latent_sde.py
mtsokol: Done!

examples/latent_sde.py
mtsokol: Done!

examples/latent_sde.py
mtsokol: Done!

lxuechen:
> Regarding taking last entry of logqp - I've tracked existing implementation and in https://github.com/google-research/torchsde/blob/master/torchsde/_core/base_solver.py#L227 logqp is appended after each step and I don't see where only last entry is retrieved.

The version on master records the `logqp` penalty accumulated on each subinterval and returns a tensor of size (T-1, batch_size), where the first dimension indexes the subinterval. The design there was so that users may explicitly weight the penalty according to chronological order if they wanted to. So if we want the vanilla `logqp` (equal weighting among different subintervals), we would sum over the first dimension. You can see that the logqp term for each subinterval is [reset to 0 at the beginning](https://github.com/google-research/torchsde/blob/9d1d65d35343be283eb6b405f7f7f0c8b0e7b106/torchsde/_core/base_solver.py#L186). 

The current version with augmentation tracks the absolute quantity, so we only need to take the end result in order to get a value consistent with what we previously had.

examples/latent_sde.py
```diff
@@ -71,23 +71,46 @@ def f(self, t, y):  # Approximate posterior drift.
...
+    def g_aug(self, t, y):
+        y = y[:, 0:1]
+        g = self.g(t, y)
+        z = torch.zeros_like(y)
+        return torch.cat([g, z], dim=1)
+
     def forward(self, ts, batch_size, eps=None):
-        eps = torch.randn(batch_size, 1).to(self.qy0_std) if eps is None else eps
+        eps = torch.cat([torch.randn(batch_size, 1).to(self.qy0_std), torch.zeros(batch_size, 1)
+                        .to(self.qy0_std)], dim=1) if eps is None else eps
```
lxuechen, line 98: This part is a little weird, since the second entry of y0 is likely not 0 (it seems to be `qy0_mean`), due to tensor broadcasting. 

Additionally, the KL divergence compute is also weird. 

This should be able to be modified by computing `eps`, `y0`, and the KL as before, but "append" y0 with zeros on the extra logqp dimension just before starting `sdeint`. 

I think this might have been a major cause in the bad performance.

examples/latent_sde.py
```diff
@@ -71,23 +71,46 @@ def f(self, t, y):  # Approximate posterior drift.
...
-                                       rtol=args.rtol, atol=args.atol)
+            zs = sdeint_adjoint(self, y0, ts, method=args.method, dt=args.dt, adaptive=args.adaptive,
+                                       rtol=args.rtol, atol=args.atol, names={'drift': 'f_aug', 'diffusion': 'g_aug'})
         else:
-            zs, logqp = sdeint(self, y0, ts, logqp=True, method=args.method, dt=args.dt, adaptive=args.adaptive,
-                               rtol=args.rtol, atol=args.atol)
+            zs = sdeint(self, y0, ts, method=args.method, dt=args.dt, adaptive=args.adaptive,
+                               rtol=args.rtol, atol=args.atol, names={'drift': 'f_aug', 'diffusion': 'g_aug'})
+        logqp = zs[-1, :, 1]
+        zs = zs[:, :, 0][:, :, None]
```
lxuechen, line 112: You could simplify this with zs = zs[:, :, 0:1].

examples/latent_sde.py
```diff
@@ -71,23 +71,46 @@ def f(self, t, y):  # Approximate posterior drift.
...
+
         logqp = logqp.sum(0).mean(0)
```
lxuechen, line 114: Does this code still run? Since `logqp` should be a tensor of (batch_size,), and you're taking the sum and mean over two axes.

examples/latent_sde.py
mtsokol: OK, so regarding `y0` it was me completely not paying attention to the code 🙄 - fixed it. 

Right now result is almost identical to `master` version (after 200 iterations):



Also after above explanation of computing subintervals and examining it once again I think now I fully understand `latent_sde.py` - so subintervals define which points we want after all from the whole solve and those here are the `x`s from `cos` training data. Then in loss we check how much prediction is off from `cos` data and we don't want to do it with just MSE because according to the paper it would cause overfitting (is it correct?), instead we check probability density function with mean at predicted values at `x`s `cos` data. 

But only fitting to the data it would make it arbitrary between data points (correct?). And here's such case after 500 iters:



Instead we want it to behave as Ornstein–Uhlenbeck process while fitting the data, so we compute through the whole solve how much our solution still behave like OU process (and regularize it that way). 

To compute the difference we use KL divergence to see how much they differ.
We start computing KL div between mean and std params of prior and posterior and then within solve - accumulate KL div for each step where KL div `dQ/dP` from Girsanov theorem will be Doléans-Dade exponential where it's `u-1/2*|u|^2`. But I can't see where `u = (f-h)/g` came from and how it would look like if prior and posterior had different diffusions, and why `u` is missing from that KL div step definiton. Do you have a link to the theory where I can look for it?

That's added to the loss, and we also decrease learning rate to make training process faster at the beginning. 

Could be `py0_mean` and `py0_logvar` be just tensors as we do not train them?

examples/latent_sde.py
mtsokol: Missed that, done!

examples/latent_sde.py
mtsokol: As comment above - removed `sum(0)`.

examples/latent_sde.py
```diff
@@ -71,24 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
         inp = torch.cat((torch.sin(t), torch.cos(t), y), dim=-1)
         return self.net(inp)
 
+    def f_aug(self, t, y):
+        y = y[:, 0:1]
+        f, g, h = self.f(t, y), self.g(t, y), self.h(t, y)
+        z = f - h
+        g = torch.where(
```
lxuechen, line 78: Could we abstract this and the next line into a helper function `_stable_division` as before in `misc`? 

The functionality of this step is clearer if we give these lines of code a name by wrapping them in a help function. 

Minor nit: You may also call the resulting value `u` to better match the derivations.

examples/latent_sde.py
```diff
@@ -71,24 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
...
+            g.abs() > 1e-7,
+            g,
+            torch.ones_like(g).fill_(1e-7) * g.sign()
+        )
+        z = z / g
+        z = .5 * (torch.norm(z, dim=1, keepdim=True) ** 2.)
```
lxuechen, line 84: I don't think the parentheses here are needed, since exponential has higher precedence than multiplication. `z = .5 * torch.norm(z, dim=1, keepdim=True) ** 2` should work.

examples/latent_sde.py
```diff
@@ -71,24 +71,47 @@ def f(self, t, y):  # Approximate posterior drift.
...
+    def g_aug(self, t, y):
+        y = y[:, 0:1]
+        g = self.g(t, y)
+        z = torch.zeros_like(y)
+        return torch.cat([g, z], dim=1)
+
     def forward(self, ts, batch_size, eps=None):
         eps = torch.randn(batch_size, 1).to(self.qy0_std) if eps is None else eps
         y0 = self.qy0_mean + eps * self.qy0_std
+        y0 = torch.cat([y0, torch.zeros(batch_size, 1).to(self.qy0_std)], dim=1)
```
lxuechen, line 99: Maybe just use `.to(y0)` to save some compute, since the property `qy0_std` is recomputed each time.

examples/latent_sde.py
lxuechen: There are easier and more intuitive ways to derive this, but for a formal derivation, you can checkout section 9.5 of our paper.

lxuechen:
Thanks for patiently addressing the issues I mentioned! I think this PR is just a few small fixes away from getting merged.

examples/latent_sde.py
mtsokol: Done!

examples/latent_sde.py
mtsokol: Yeah sure, done!

examples/latent_sde.py
mtsokol: Done!

mtsokol:
@lxuechen All done! 
Last question that I forgot to ask: Why in this example we encode NN input as (sin(t), cos(t), y) instead of just (t, y)? 
As also with such formula second input is determined by the first one which seems redundant.

torchsde/_core/misc.py
```diff
@@ -98,3 +98,12 @@ def flat_to_shape(tensor, shapes, length=()):
         tensor_list.append(tensor[..., total:next_total].view((*length, *shape)))
         total = next_total
     return tensor_list
+
+
+def _stable_division(x, y, epsilon=1e-7):
+    y = torch.where(
+        y.abs() > epsilon,
+        y,
+        torch.ones_like(y).fill_(epsilon) * y.sign()
```
patrick-kidger, line 107: Quick comment: can this be `torch.full_like(y, epsilon * y.sign())` instead?
(Untested, might have the syntax slightly wrong.)

mtsokol: Sure!
APi says that second arg should be a number. This works for me: `torch.full_like(y, epsilon) * y.sign()`

examples/latent_sde.py
```diff
@@ -28,6 +28,7 @@
 
 from examples import utils
 from torchsde import sdeint, sdeint_adjoint, SDEIto, BrownianInterval
+from torchsde._core.misc import _stable_division
```
lxuechen, line 31: Let's not put `_stable_division` in core, but directly in this file.

lxuechen:
> @lxuechen All done!
> Last question that I forgot to ask: Why in this example do we encode NN input as (sin(t), cos(t), y) instead of just (t, y)?
> As also with such formula second input is determined by the first one which seems redundant.

This is to mimic positional encoding in transformers. Also it's incorrect to say that the value of sine determines the value of cosine, e.g. sin(pi/4) == sin(3pi/4), but cos(pi/4) != cos(3pi/4). This is elementary trigonometric.

mtsokol:
All done!

> > @lxuechen All done!
> > Last question that I forgot to ask: Why in this example do we encode NN input as (sin(t), cos(t), y) instead of just (t, y)?
> > As also with such formula second input is determined by the first one which seems redundant.
> 
> This is to mimic positional encoding in transformers. Also it's incorrect to say that the value of sine determines the value of cosine, e.g. sin(pi/4) == sin(3pi/4), but cos(pi/4) != cos(3pi/4). This is elementary trigonometric.

I was thinking about something completely different while writing this, not even looking at what functions are used, let's say because of current late night hour _(embarrassment)_

examples/latent_sde.py
mtsokol: Sure!

torchsde/_core/misc.py
mtsokol: Done!

lxuechen:
> All done!
> 
> > > @lxuechen All done!
> > > Last question that I forgot to ask: Why in this example do we encode NN input as (sin(t), cos(t), y) instead of just (t, y)?
> > > As also with such formula second input is determined by the first one which seems redundant.
> > 
> > 
> > This is to mimic positional encoding in transformers. Also it's incorrect to say that the value of sine determines the value of cosine, e.g. sin(pi/4) == sin(3pi/4), but cos(pi/4) != cos(3pi/4). This is elementary trigonometric.
> 
> I was thinking about something completely different while writing this, not even looking at what functions are used, let's say because of current late night hour _(embarrassment)_

No worries! I understand that might happen at times.

mtsokol:
Thanks for all the assistance once again! This or the following week I will try to do next small thing from milestone list (also started simple benchmark which I was asking earlier, will see how it goes).

(I thought about turning open source contributions to some simple master thesis to be able to spend more time on this but unfortunately I haven't found anyone interested in supervising it at uni for coming year).

---

Also I've got a question related to constraints of gradient computation (please let me know if it's inappropriate to ask here!)
Some time ago I was learning about [Physics Informed Neural Networks](https://maziarraissi.github.io/PINNs/) idea for solving PDEs from it's original paper and Tensorflow source code. I experimented with a small tweak - instead of learning whole solution `u(x,t)` with NN, I thought about introducing bspline base functions for solution like in FEM and using NN to learn it's coefficients. I've created full prototype:
https://github.com/pierremtb/PINNs-TF2.0/pull/3
But eventually it just pull all coefficients close to zero and doesn't recreate desired shape. 

As there's a heat equation example and PINNs use gradient descent it eventually computes higher order derivatives of Tensorflow's control flows like loops and ifs (used for choosing correct bsplines). (I also tried learning coefficients straight from GD but also failed)

So do you think that computing gradient of such complex logic might be unfeasible to acquire solution or there's no reason to say that and it's rather implemented incorrectly?

lxuechen:
Sorry for the late response, as I've been incredibly busy lately. 

Obviously, I'm not an expert on the specific models you described. What I may potentially comment on in an educated manner is the part about differentiating through control flow and obtaining high-order derivatives. 

Since both TF *eager-mode* and Pytorch use tape-based systems, taking gradients through control flow shouldn't be a problem, even if that control flow is pure Python code (Python if-else, for, while statements). Obviously, things get tricky when you start with TF graph-mode and jitting/scripting PyTorch code, and things typically break without there being potential simple fixes. Obviously, this is only coming from my limited experience.

Second-order gradients typically aren't a problem either, if the specific computation can be grouped into Hessian-vector products, which can then be computed using just vector-Jacobian products. I haven't seen examples of taking gradients beyond the second-order in the ML literature so far.

PR #49, opened 2020-09-06
Added double adjoint

patrick-kidger:
Just creating a draft PR to let you know I'm working on double-backward through the adjoint.

In informal tests it seems to work as expected. The only thing left to do is write some formal tests. I'll put them in `test_adjoint.py` and switch that file over to pytest while I'm at it.

### DIFF
--- a/torchsde/_core/adjoint.py
+++ b/torchsde/_core/adjoint.py
@@ -41,7 +41,7 @@ class _SdeintAdjointMethod(torch.autograd.Function):
 
         ys = sdeint.integrate(
             sde=sde,
-            y0=y0,
+            y0=y0.detach(),  # This .detach() is VERY IMPORTANT. See adjoint_sde.py::AdjointSDE._unpack_y_aug.
             ts=ts,
             bm=bm,
             method=method,
@@ -75,19 +75,10 @@ class _SdeintAdjointMethod(torch.autograd.Function):
 
         for i in range(ys.size(0) - 1, 0, -1):
             aug_state = misc.flatten(aug_state)
-            aug_state = sdeint.integrate(
-                sde=adjoint_sde,
-                y0=aug_state,
-                ts=torch.stack([-ts[i], -ts[i - 1]]),
-                bm=reverse_bm,
-                method=adjoint_method,
-                dt=dt,
-                adaptive=adjoint_adaptive,
-                rtol=adjoint_rtol,
-                atol=adjoint_atol,
-                dt_min=dt_min,
-                options=adjoint_options
-            )
+            aug_state = _SdeintAdjointMethod.apply(adjoint_sde, torch.stack([-ts[i], -ts[i - 1]]), dt, reverse_bm,
+                                                   adjoint_method, adjoint_method, adjoint_adaptive, adjoint_adaptive,
+                                                   adjoint_rtol, adjoint_rtol, adjoint_atol, adjoint_atol, dt_min,
+                                                   adjoint_options, adjoint_options, aug_state, *params)
             aug_state = misc.flat_to_shape(aug_state[1], shapes)  # Unpack the state at time -ts[i - 1].
             aug_state[0] = ys[i - 1]
             aug_state[1] = aug_state[1] + grad_ys[i - 1]
@@ -106,10 +97,10 @@ def sdeint_adjoint(sde: nn.Module,
                    dt: Optional[Scalar] = 1e-3,
                    adaptive: Optional[bool] = False,
                    adjoint_adaptive: Optional[bool] = False,
-                   rtol: Optional[float] = 1e-5,
-                   adjoint_rtol: Optional[float] = 1e-5,
-                   atol: Optional[float] = 1e-4,
-                   adjoint_atol: Optional[float] = 1e-4,
+                   rtol: Optional[Scalar] = 1e-5,
+                   adjoint_rtol: Optional[Scalar] = 1e-5,
+                   atol: Optional[Scalar] = 1e-4,
+                   adjoint_atol: Optional[Scalar] = 1e-4,
                    dt_min: Optional[Scalar] = 1e-5,
                    options: Optional[Dict[str, Any]] = None,
                    adjoint_options: Optional[Dict[str, Any]] = None,
@@ -167,6 +158,8 @@ def sdeint_adjoint(sde: nn.Module,
         raise ValueError("`sde` is required to be an instance of nn.Module.")
 
     sde, y0, ts, bm = sdeint.check_contract(sde, y0, ts, bm, method, names)
+    misc.assert_no_grad(['ts', 'dt', 'rtol', 'adjoint_rtol', 'atol', 'adjoint_atol', 'dt_min'],
+                        [ts, dt, rtol, adjoint_rtol, atol, adjoint_atol, dt_min])
     adjoint_method = _select_default_adjoint_method(sde, adjoint_method)
     params = filter(lambda x: x.requires_grad, sde.parameters())
 
--- a/torchsde/_core/adjoint.py
+++ b/torchsde/_core/adjoint.py
@@ -169,10 +169,12 @@ def sdeint_adjoint(sde: nn.Module,
             or `sde` is missing required methods.
 
     Note:
-        Double-backward is supported, and will use the adjoint method to compute the gradient of the adjoint. (i.e.
-        rather than backpropagating through the numerical solver used for the adjoint.) The same `adjoint_method`,
-        `adjoint_adaptive`, `adjoint_rtol, `adjoint_atol`, `adjoint_options` will be used for the second-order adjoint
-        as is used for the first-order adjoint.
+        Double-backward is supported, and will use the adjoint method to compute
+        the gradient of the adjoint. (i.e. rather than backpropagating through
+        the numerical solver used for the adjoint.) The same `adjoint_method`,
+        `adjoint_adaptive`, `adjoint_rtol, `adjoint_atol`, `adjoint_options`
+        will be used for the second-order adjoint as is used for the first-order
+        adjoint.
     """
     misc.handle_unused_kwargs(unused_kwargs, msg="`sdeint_adjoint`")
     del unused_kwargs
--- a/torchsde/_core/adjoint_sde.py
+++ b/torchsde/_core/adjoint_sde.py
@@ -58,32 +58,72 @@ class AdjointSDE(base_sde.BaseSDE):
             NOISE_TYPES.diagonal: self.gdg_prod_diagonal,
         }.get(sde.noise_type, self.gdg_prod_default)
 
-    def _unpack_y_aug(self, y_aug):
+    def _unpack_y_aug(self, t, y_aug, v):
+        # These leaf checks are very important.
+        # _unpack_y_aug is used where we want to compute:
+        # ```
+        # with torch.enable_grad():
+        #     s = some_function(y)
+        #     torch.autograd.grad(s, [y] + params, ...)
+        # ```
+        # where `some_function` implicitly depends on `params`.
+        # However if y has history of its own then in principle it could _also_ depend upon `params`, and this call to
+        # `grad` will go all the way back to that.
+        # To avoid this, we require that y (and the other tensors) be leaf tensors.
+        #
+        # Note that this is also the reason for the `y0.detach()` in adjoint.py::_SdeintAdjointMethod.forward. Unless
+        # we detach then y0 may have a history and these checks will fail. Note that this isn't then producing the wrong
+        # gradients when backprop'ing as usual because `torch.autograd.Function.forward` has an implicit
+        # `torch.no_grad()` guard, i.e. any history doesn't affect the usual gradient computations anyway. But it does
+        # still get detected if we call `torch.autograd.grad` ourselves, which is explicitly what we don't want as per
+        # the previous paragraph.
+        assert t.is_leaf, "Internal error: please report a bug to torchsde"
+        assert y_aug.is_leaf, "Internal error: please report a bug to torchsde"
+        assert v.is_leaf, "Internal error: please report a bug to torchsde"
+
+        # This determines whether or not we will be able to backpropagate through the calling function (that is calling
+        # _unpack_y_aug). Simply, we should be able to if and only if (a) gradients are enabled and (b) the input
+        # arguments require gradient.
+        # Note that we don't fix this to True, because that implies building computational graphs, which will consume
+        # additional memory that will be unnecessary if we don't need to backpropagate.
+        requires_grad = torch.is_grad_enabled() and (t.requires_grad or y_aug.requires_grad or v.requires_grad)
+
         y, adj_y = misc.flat_to_shape(y_aug, self._shapes[:2])
-        return y.detach().requires_grad_(True), adj_y.detach().requires_grad_(True)
+
+        # To support the later differentiation wrt y, we set it to require_grad if it doesn't already.
+        if not y.requires_grad:
+            y = y.detach().requires_grad_(True)
+        return y, adj_y, requires_grad
 
     ########################################
     #                  f                   #
     ########################################
 
     def f_uncorrected(self, t, y_aug):  # For Ito additive and Stratonovich.
+        y, adj_y, requires_grad = self._unpack_y_aug(t, y_aug, v=t)  # just use t as a dummy `v`
         with torch.enable_grad():
-            y, adj_y = self._unpack_y_aug(y_aug)
             f = self._base_sde.f(-t, y)
             vjp_y_and_params = misc.vjp(
                 outputs=f,
                 inputs=[y] + self._params,
                 grad_outputs=adj_y,
-                allow_unused=True
+                allow_unused=True,
+                create_graph=requires_grad
             )
+            if not requires_grad:
+                # We had to build a computational graph to be able to compute the above vjp.
+                # However, if we don't require_grad then we don't need to backprop through this function, so we should
+                # delete the computational graph to avoid a memory leak. (Which for example would keep the local
+                # variable `y` in memory: f->grad_fn->...->AccumulatedGrad->y.)
+                f = f.detach()
         return misc.flatten((-f, *vjp_y_and_params))
 
     def f_corrected_default(self, t, y_aug):  # For Ito general/scalar.
         raise NotImplementedError
 
     def f_corrected_diagonal(self, t, y_aug):  # For Ito diagonal.
+        y, adj_y, requires_grad = self._unpack_y_aug(t, y_aug, v=t)  # just use t as a dummy `v`
         with torch.enable_grad():
-            y, adj_y = self._unpack_y_aug(y_aug)
             g = self._base_sde.g(-t, y)
             g_dg_vjp, = misc.vjp(
                 outputs=g,
@@ -99,21 +139,27 @@ class AdjointSDE(base_sde.BaseSDE):
                 inputs=[y] + self._params,
                 grad_outputs=adj_y,
                 allow_unused=True,
-                create_graph=True
+                retain_graph=True,
+                create_graph=requires_grad
             )
+            if not requires_grad:
+                # See corresponding note in f_uncorrected.
+                f = f.detach()
             # Convert the adjoint Stratonovich SDE to Itô form.
             a_dg_vjp, = misc.vjp(
                 outputs=g,
                 inputs=y,
                 grad_outputs=adj_y,
                 allow_unused=True,
-                create_graph=True
+                retain_graph=True,
+                create_graph=requires_grad
             )
             extra_vjp_y_and_params = misc.vjp(
                 outputs=g,
                 inputs=[y] + self._params,
                 grad_outputs=a_dg_vjp,
                 allow_unused=True,
+                create_graph=requires_grad
             )
             vjp_y_and_params = misc.seq_add(vjp_y_and_params, extra_vjp_y_and_params)
         return misc.flatten((-f, *vjp_y_and_params))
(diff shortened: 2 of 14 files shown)


torchsde/_core/adjoint_sde.py
```diff
@@ -58,32 +58,72 @@ def __init__(self,
...
+        if not y.requires_grad:
+            y = y.detach().requires_grad_(True)
+        return y, adj_y, requires_grad
 
     ########################################
     #                  f                   #
     ########################################
 
     def f_uncorrected(self, t, y_aug):  # For Ito additive and Stratonovich.
+        y, adj_y, requires_grad = self._unpack_y_aug(t, y_aug, v=t)  # just use t as a dummy `v`
```
lxuechen, line 103: It seems a bit weird that `requires_grad` is returned by `_unpack_y_aug`, as I think the method should solely be responsible for unpacking `y_aug` as its name says. 

I think we should be able to define `requires_grad` before calling `_unpack_y_aug`. This perhaps will also read better.

torchsde/_core/adjoint_sde.py
```diff
@@ -58,32 +58,72 @@ def __init__(self,
...
+        # `grad` will go all the way back to that.
+        # To avoid this, we require that y (and the other tensors) be leaf tensors.
+        #
+        # Note that this is also the reason for the `y0.detach()` in adjoint.py::_SdeintAdjointMethod.forward. Unless
+        # we detach then y0 may have a history and these checks will fail. Note that this isn't then producing the wrong
+        # gradients when backprop'ing as usual because `torch.autograd.Function.forward` has an implicit
+        # `torch.no_grad()` guard, i.e. any history doesn't affect the usual gradient computations anyway. But it does
+        # still get detected if we call `torch.autograd.grad` ourselves, which is explicitly what we don't want as per
+        # the previous paragraph.
+        assert t.is_leaf, "Internal error: please report a bug to torchsde"
```
lxuechen, line 80: Do we have an application/use-case where we need to call `torch.autograd.grad` where `t` is the input?

torchsde/_core/adjoint_sde.py
```diff
@@ -58,32 +58,72 @@ def __init__(self,
             NOISE_TYPES.diagonal: self.gdg_prod_diagonal,
         }.get(sde.noise_type, self.gdg_prod_default)
 
-    def _unpack_y_aug(self, y_aug):
+    def _unpack_y_aug(self, t, y_aug, v):
+        # These leaf checks are very important.
```
lxuechen, line 62: Could we shorten this comment for a bit? It's becoming design-doc-like. The crux should be that this prevents `grad` from going into `y` towards `params`, right?

torchsde/_core/adjoint_sde.py
```diff
@@ -58,32 +58,72 @@ def __init__(self,
...
+        assert t.is_leaf, "Internal error: please report a bug to torchsde"
+        assert y_aug.is_leaf, "Internal error: please report a bug to torchsde"
+        assert v.is_leaf, "Internal error: please report a bug to torchsde"
+
+        # This determines whether or not we will be able to backpropagate through the calling function (that is calling
+        # _unpack_y_aug). Simply, we should be able to if and only if (a) gradients are enabled and (b) the input
+        # arguments require gradient.
+        # Note that we don't fix this to True, because that implies building computational graphs, which will consume
+        # additional memory that will be unnecessary if we don't need to backpropagate.
+        requires_grad = torch.is_grad_enabled() and (t.requires_grad or y_aug.requires_grad or v.requires_grad)
```
lxuechen, line 89: Do we have an application/use-case where we need to call `torch.autograd.grad` where t is the input?

torchsde/_core/base_sde.py
```diff
@@ -91,10 +91,10 @@ def gdg_prod_default(self, t, y, v):
         return vg_dg_vjp
 
     def gdg_prod_diagonal(self, t, y, v):
-        requires_grad = torch.is_grad_enabled()
+        requires_grad = torch.is_grad_enabled() and (t.requires_grad or y.requires_grad or v.requires_grad)
```
lxuechen, line 94: Do we have an application/use-case where we need to call torch.autograd.grad where t is the input?

torchsde/_core/adjoint_sde.py
lxuechen, line 103: It seems a bit weird to use `t` as dummy.

torchsde/_core/adjoint_sde.py
```diff
@@ -58,32 +58,72 @@ def __init__(self,
...
+        #
+        # Note that this is also the reason for the `y0.detach()` in adjoint.py::_SdeintAdjointMethod.forward. Unless
+        # we detach then y0 may have a history and these checks will fail. Note that this isn't then producing the wrong
+        # gradients when backprop'ing as usual because `torch.autograd.Function.forward` has an implicit
+        # `torch.no_grad()` guard, i.e. any history doesn't affect the usual gradient computations anyway. But it does
+        # still get detected if we call `torch.autograd.grad` ourselves, which is explicitly what we don't want as per
+        # the previous paragraph.
+        assert t.is_leaf, "Internal error: please report a bug to torchsde"
+        assert y_aug.is_leaf, "Internal error: please report a bug to torchsde"
+        assert v.is_leaf, "Internal error: please report a bug to torchsde"
```
lxuechen, line 82: I think if we assume `v` comes from the Brownian motion, then it's also safe to assume that it's a leaf, since tensors generated by BM doesn't require gradients, and so is the returning tensor of any function of such tensors.

lxuechen:
Thanks for pulling off a draft!

Some general comments

- Do we need to include `t` and `v` in `_unpack_y_aug`. My understanding is that we shouldn't unless 1) there's an explicit use-case where we need to take derivative w.r.t. `t`, and 2) there's an explicit use-case where `v` is not from the Brownian motion but some other tensor that requires grad. I haven't seen cases for 1), but I'm not entirely sure there isn't such a case. I don't think we'll ever see cases for 2). 

- It seems to me that since `_SdeintAdjointMethod.apply` is called in `backward`, the context returned by `torch.is_grad_enabled` in the adjoint functions (e.g. `f_uncorr`) should be `False` due to the implicit context setting with `forward` of `torch.autograd.Function`. 

- The current paradigm is adjoints-through-adjoints, and doesn't support naive backprop through adjoints. We should probably document this, even though this may seem trivial to us. 

> In informal tests it seems to work as expected.

What are the informal tests? Are there finite-difference tests? Or did we directly use `torch.autograd.gradgradcheck`?  Are there any companion tests and does the old tests still pass? 

Hopefully asking all of these question at once doesn't seem too annoying, but I just want to gain a better understanding of what's happening here. Also, we don't have to do everything at once or address all of the questions above at once, but it's good to know what works and what doesn't so that we can place meaningful TODOs.

torchsde/_core/adjoint.py
```diff
@@ -75,19 +75,10 @@ def backward(ctx, grad_ys):  # noqa
...
-                bm=reverse_bm,
-                method=adjoint_method,
-                dt=dt,
-                adaptive=adjoint_adaptive,
-                rtol=adjoint_rtol,
-                atol=adjoint_atol,
-                dt_min=dt_min,
-                options=adjoint_options
-            )
+            aug_state = _SdeintAdjointMethod.apply(adjoint_sde, torch.stack([-ts[i], -ts[i - 1]]), dt, reverse_bm,
```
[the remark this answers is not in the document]
patrick-kidger: Done.

torchsde/_core/adjoint_sde.py
patrick-kidger: So it's there because the decision about whether to call `.detach().requires_grad_(True)` on `y` is also part of `_unpack_y_aug`'s responsibility, and it felt neatest to keep similarly-motivated pieces together.

Renaming `_unpack_y_aug` is probably simplest here.

torchsde/_core/adjoint_sde.py
patrick-kidger: I don't think so, but that's not what matters - if `t` has history (input or not) then things could go wrong.

torchsde/_core/adjoint_sde.py
patrick-kidger: I'm trying to write for the developer five years from now who's forgotten everything about this. What's the issue with long comments?

torchsde/_core/base_sde.py
patrick-kidger: See below.

torchsde/_core/adjoint_sde.py
patrick-kidger: See below.

torchsde/_core/adjoint_sde.py
patrick-kidger: Regarding this and the points about `t`: AFAIK both `t` and `v` should never require gradients in the current set-up.

But that's a detail that's beyond `AdjointSDE`'s scope. `AdjointSDE` requires that these conditions are satisfied, so it should enforce them. Practically: if that detail ever changes down the line then I'd prefer to know about it without things silently going wrong.

torchsde/_core/adjoint_sde.py
patrick-kidger: It seemed neatest to create a single `_unpack_y_aug` for both cases.

patrick-kidger:
- Including `t` and `v`: AFAIK both `t` and `v` should never require gradients in the current set-up.

But that's a detail that's beyond `AdjointSDE`'s scope. `AdjointSDE` requires that certain conditions are satisfied, so it should enforce them. Practically: if that detail ever changes down the line then I'd prefer to know about it without things silently going wrong. The logic here can get quite hairy when you start getting into adjoints-of-adjoints.

- implicit contexts: this isn't true in the double-adjoint case, when `f_uncorrected` of the double adjoint sets `enable_grad` and then calls `f_uncorrected` of the first adjoint.

- Naive backprop: yeah, it should probably be added to the docstring for `sdeint_adjoint`. I'll do that.

Old single-adjoint tests still pass.

lxuechen:
> Including t and v: AFAIK both t and v should never require gradients in the current set-up.

Per current usage, I can't really come up with a case where `t` and `v` are not leaf variables, so I would prefer to not let `_get_state` check these. I think I would leave some comments here if we're really feeling unsure. 

> implicit contexts: this isn't true in the double-adjoint case, when f_uncorrected of the double adjoint sets enable_grad and then calls f_uncorrected of the first adjoint.

You're right, thanks for explaining. 

> Naive backprop: yeah, it should probably be added to the docstring for sdeint_adjoint. I'll do that.

Thanks for the fix for documentation!

Additionally, could we do a numerical test? Ideally, just replacing the `torch.autograd.gradcheck` [here](https://github.com/google-research/torchsde/blob/0bfddc449c2c9d3e21b24c98dd0010e8e08631f0/tests/test_strat.py#L113) with `torch.autograd.gradgradcheck` (and slightly modifying the surrounding) should tell us something.

torchsde/_core/adjoint_sde.py
lxuechen: The reason I'm asking for some simplifications is to prevent the code from getting too bulky to read (I personally had to parse this text for some time!). 

I certainly think this comment section could be shortened a little, but if you insist on including this unmodified, then I'd recommend wrapping this in the docstring, in case we'd like to auto-generate documentation webpages based on docstrings, and that in the rare occasion we'd like to call `help` on this function.

torchsde/_core/adjoint_sde.py
```diff
@@ -58,32 +58,72 @@ def __init__(self,
             NOISE_TYPES.diagonal: self.gdg_prod_diagonal,
         }.get(sde.noise_type, self.gdg_prod_default)
 
-    def _unpack_y_aug(self, y_aug):
+    def _get_state(self, t, y_aug, v):
```
lxuechen, line 61: The name `_get_state` is not very meaningful. I'd either change this to something more meaningful, or include a one-liner docstring header explaining its purposes.

torchsde/_core/adjoint_sde.py
lxuechen: I don't think `t` should require gradients at any point, meaning it shouldn't have a history for gradient computation purposes. Am I correct in saying this?

torchsde/_core/adjoint_sde.py
```diff
@@ -99,21 +139,27 @@ def f_corrected_diagonal(self, t, y_aug):  # For Ito diagonal.
                 inputs=[y] + self._params,
                 grad_outputs=adj_y,
                 allow_unused=True,
-                create_graph=True
+                retain_graph=True,
+                create_graph=requires_grad
             )
+            if not requires_grad:
```
lxuechen, line 145: For the sake of consistency, do you think it'd be better to put this detach statement near the bottom just before the `return ...` statement?

torchsde/_core/base_sde.py
```diff
@@ -66,21 +66,21 @@ def __init__(self, sde, fast_dg_ga_jvp_column_sum=False):
...
+        return misc.batch_mvp(self.g(t, y), v)
 
     ########################################
     #               gdg_prod               #
     ########################################
 
     # Computes: sum_{j, l} g_{j, l} d g_{j, l} d x_i v_l.
     def gdg_prod_default(self, t, y, v):
-        requires_grad = torch.is_grad_enabled()
+        requires_grad = torch.is_grad_enabled() and (t.requires_grad or y.requires_grad or v.requires_grad)
```
lxuechen, line 80: I'm a little worried about this line. Consider the scenario where we set `y0` to not require gradients. By current standards, neither `t` nor `v` should require gradients. So in the backprop-through-solver setting, `requires_grad` evaluates to False. 

The problematic scenario is then solving an SDE with parameters that require gradients. Since `requires_grad` is False, we don't create the extra graph backpropping through `vg_dg_vjp` towards the parameters. However, in the ideal situation, we should still backprop through this node so that our models get better gradients. 

Overall, I think this possible failure case indicates that we should have a numerical test for backprop-through-solver comparing against finite differences.

torchsde/_core/adjoint_sde.py
patrick-kidger: Added docstring, and trimmed the comment a little.

torchsde/_core/adjoint_sde.py
patrick-kidger: Done.

torchsde/_core/base_sde.py
patrick-kidger: I think it's worth noting that we currently don't (a) backprop-through-the-solver with the adjoint, nor do we (b) differentiate wrt parameters without also differentiating wrt y, so as the intersection of these two null events we at least shouldn't be any running into any issues with the current codebase.

That said, this is a subtle point that we'd probably never catch in the future if this scenario does come up, so I agree that I'd prefer to do something about it now.

I'm not completely sure about a fix.
The current `requires_grad` is meant to correspond to "if I hadn't forced `y` to have a gradient, would `some_func(t, y, v)` require gradient?"
I'm -1 on any approach that just fixes it to `True`, as else we'll be leaking graphs from the function.
I think my proposed fix would be to:
- If `y` requires gradient, set `requires_grad` to True. (As before.)
- If `y` doesn't require gradient, then we're going to be forcing it to require gradient. Then we call `some_func(t, y, v)` as before. Prior to calling `vjp` we manually traverse its autograd history (via `.grad_fn`, `.next_functions`) and set `requires_grad = not (is y the only leaf)`.

Not a huge fan of that but I don't have any better ideas. Thoughts?

torchsde/_core/adjoint_sde.py
patrick-kidger: I've shortened the comment to the extent that I think's reasonable.
I deliberately didn't put it in the docstring because it's really just about the `.is_leaf` checks rather than the function as a whole.

torchsde/_core/adjoint_sde.py
patrick-kidger: See my main-thread comment on this.

patrick-kidger:
I agree that every current case involves `t` and `v` being leaf variables already. (Moreover non-gradient-requiring variables.)
I'm quite strongly in favour of maintaining the checks though, for what I think is the same reason that you brought up the discussion on `requires_grad`: we might not be doing it now, but it's an implicit assumption that will silently do the wrong thing if we ever change that assumption in the future.

Regarding tests - I feel like you think I'm a bit cavalier about tests! Don't worry, I appreciate their importance for any software project; I just tend to put them in a bit later than you.
For this PR specifically, it sounds like you're offering to write the tests? If you get time and are happy to do it, that would be great.

torchsde/_core/base_sde.py
lxuechen: > I think it's worth noting that we currently don't (a) backprop-through-the-solver with the adjoint, nor do we (b) differentiate wrt parameters without also differentiating wrt y, so we at least shouldn't be any running into any issues with the current codebase.

I don't think the line I'm specifically referring to has much to do with adjoints. `gdg_prod` is part of `ForwardSDE`, so in the backprop through solver case, it would get called when we choose, say, the Milstein method. 

More specifically, I think this will cause a problem in the first step of sdeint (w/o adjoints), where `y0` doesn't require gradient. The later steps shouldn't be affected by too much, since the typical `y` beyond `y0` should require gradients due to being sums of tensors that require gradient with tensors that don't.

patrick-kidger: Oh sorry, you're commenting on base_sde rather than adjoint_sde. You're right, this would break Milstein.

I think the analysis for the forward case is the same as the backward case that I actually discussed, though, so I think I'd end up proposing the same fix. I'll see what you say.

Unrelatedly, I've updated `test_adjoint.py` to pytest. Scalar noise+Ito SDE seems to be failing the tests btw. (It wasn't tested previously.) I've not dug into why that is - I know this isn't a supported case on master but I thought it was in now?

lxuechen: I think the issue I mentioned is mostly about when `sde.parameters()` is non-empty. So I would perhaps add a literal that checks whether `ForwardSDE` has any parameters that require gradients. 

I'll fix the Scalar noise + Ito case.

lxuechen:
> I agree that every current case involves `t` and `v` being leaf variables already. (Moreover non-gradient-requiring variables.)
> I'm quite strongly in favour of maintaining the checks though, for what I think is the same reason that you brought up the discussion on `requires_grad`: we might not be doing it now, but it's an implicit assumption that will silently do the wrong thing if we ever change that assumption in the future.
> 
> Regarding tests - I feel like you think I'm a bit cavalier about tests! Don't worry, I appreciate their importance for any software project; I just tend to put them in a bit later than you.
> For this PR specifically, it sounds like you're offering to write the tests? If you get time and are happy to do it, that would be great.

I appreciate the detailed thoughts, though I still believe there's a concrete difference between the issue I mentioned and the argument about `t` and `v`. I think we currently do have a use case of backprop through solver, and the issue I mentioned could affect this behavior. On the other hand, I couldn't really come up with a case where gradients wrt t and v need to be taken.

torchsde/_core/base_sde.py
patrick-kidger: So that relies on the implicit assumption that everything going into the function is a parameter. That needn't be the case, c.f. our discussion a bit back on contextualisation.

patrick-kidger:
Hmm I don't quite follow. Gradients wrt `t` and `v` are a separate issue to this breaking Milstein. (If that's what you're saying then I agree!)
I agree that we never need gradients wrt `t` or `v`, I'm just quite assert-happy as a way of making sure things aren't silently misbehaving.
I agree that this PR introduces an issue wrt Milstein (whilst fixing the issue wrt leaking graphs).

Reflecting on the main (gradient) issue, I am actually inclined to just set `requires_grad = torch.is_grad_enabled()` and leave it at that. This does leak graphs, but I don't think the overhead should be too large. I think the alternate fix, or manually checking the graph, is probably a bit too magic.

lxuechen:
> Hmm I don't quite follow. Gradients wrt `t` and `v` are a separate issue to this breaking Milstein. (If that's what you're saying then I agree!)

Yes, I'm trying to argue that these are separate situations. 

> I agree that we never need gradients wrt `t` or `v`, I'm just quite assert-happy as a way of making sure things aren't silently misbehaving.

I still don't think it's necessary to check `t` or `v`, as I'm unable to come up with any likely scenario where these variables would require grad. Given that this check also adds complexity to the code and a tiny overhead, I'm not sure of the advantage. 

> I agree that this PR introduces an issue wrt Milstein (whilst fixing the issue wrt leaking graphs).
> 
> Reflecting on the main (gradient) issue, I am actually inclined to just set `requires_grad = torch.is_grad_enabled()` and leave it at that. This does leak graphs, but I don't think the overhead should be too large. I think the alternate fix, or manually checking the graph, is probably a bit too magic.

I agree with this fix, i.e. just leaving `requires_grad = torch.is_grad_enabled()`. 

Side note, I think this PR might be ready to be converted to review-mode from draft mode modulo the `t` `v` grad check discussion.

patrick-kidger:
Changed `requires_grad`.

Regarding the asserts: as I understand it, you want to remove the `assert t.is_leaf` and `assert v.is_leaf` lines from `adjoint_sde.py`? I think it's important that `AdjointSDE` (or indeed any other library component) is self-contained; explicitly checked assumptions are better than implicit assumptions, especially in cases like this where a failure of that assumption will fail silently rather than fail loudly.

I can't think of a use case for `t` or `v` needing gradients either - but that could change not just because of some unforeseen feature, but because of a bug. (And the cost/complexity of two asserts is very low.)

lxuechen:
Ok, let's leave the checks as they are for now. Happy to have this merged and thanks again for the great work in getting the adjoint-adjoints working!

PR #50, opened 2020-09-07
Add falling back imports for brownian_lib

dboyliao:
Add falling back import statements for `brownian_lib` module.
Or tests may fail to run due to `AttributeError` or `ImportError`

### DIFF
--- a/torchsde/brownian_lib/__init__.py
+++ b/torchsde/brownian_lib/__init__.py
@@ -21,4 +21,6 @@ try:
     BrownianPath.__init__.__annotations__ = {}
     BrownianTree.__init__.__annotations__ = {}
 except Exception:  # noqa
-    warnings.warn('Failed to import `torchsde._brownian_lib`; falling back to `torchsde._brownian`.')
+    warnings.warn('Failed to import `torchsde.brownian_lib`; falling back to `torchsde._brownian`.')
+    from .._brownian import BrownianPath
+    from .._brownian import BrownianTree


dboyliao:
Screen Shot 2020-09-07 at 4 27 26 PM

dboyliao:
@googlebot I fixed it.

patrick-kidger:
Googlebot doesn't seem convinced! Did you enter your GitHub user name when you signed the CLA?

In any case - this is only an issue for running the tests, correct? In which case I'd suggest that probably the better thing to do would be to modify the tests to skip `brownian_lib` if it doesn't exist. (Also I'd suggest working from the `dev` branch rather than the `master` branch.)

lxuechen:
So the folder `brownian_lib` is to explicitly differentiate between the Python implementation and C++ implementation. 

It doesn't really make sense to let `brownian_lib` import the same classes. I'd probably modify the tests with something like
```python
if not hasattr(torchsde.brownian_lib, "BrownianPath"):
...
```
or 
```python
try:
    torchsde.brownian_lib.BrownianPath
except AttributeError:
...
```
Also, I appreciate the effort in sending in a PR, but I'd recommend opening up an issue first. It doesn't quite make sense to send in a PR before initiating any type of discussion.

dboyliao:
@patrick-kidger Yes, I managed to get tests work with adding imports.
I inspect the code of the tests and find out that `BrownianPath` and `BrownianTree` from both modules should work compatibly.
That's why I come up with fix.
I check the link googlebot gave me and it looks like I did sign the CLA before.
Not sure why it does not work.

@lxuechen You are right, I should go for issue tracker first.
Sorry for rushing an PR without discussion.
Will close this one and open up an issue instead.
Thanks for your suggestions.

PR #57, opened 2020-09-17
Dev adjoint double test

lxuechen:
Sorry that this test took so long. For long time horizons, the finite differences seem to blow up. Though apart from that particular instance, the tests pass. I also needed to use small dimensions (e.g. batch size, d, m) so that the finite differences finish under reasonable time.

### DIFF
--- a/tests/test_strat.py
+++ b/tests/test_strat.py
@@ -17,22 +17,27 @@
 This should be eventually refactored and the file should be removed.
 """
 
+import copy
 import time
 
 import torch
+from torch import nn
 
 from torchsde import sdeint_adjoint, BrownianInterval
+from torchsde._core import misc  # noqa
 from torchsde._core.base_sde import ForwardSDE  # noqa
 from torchsde.settings import SDE_TYPES
-from .problems import Ex4
+from torchsde.types import Callable
+from . import problems
 
+torch.set_printoptions(precision=10)
 torch.manual_seed(1147481649)
 torch.set_default_dtype(torch.float64)
 cpu, gpu = torch.device('cpu'), torch.device('cuda')
 device = gpu if torch.cuda.is_available() else cpu
 dtype = torch.get_default_dtype()
-batch_size, d, m = 1, 2, 3
-ts = torch.tensor([0.0, 0.2, 0.4], device=device)
+batch_size, d, m = 5, 2, 3
+ts = torch.tensor([0.0, 0.2, 1.0, 1.5], device=device)
 t0, t1 = ts[0], ts[-1]
 y0 = torch.full((batch_size, d), 0.1, device=device)
 
@@ -69,7 +74,7 @@ def _make_inputs():
     y = torch.randn(batch_size, d, device=device)
     a = torch.randn(batch_size, m, m, device=device)
     a = a - a.transpose(1, 2)  # Anti-symmetric.
-    sde = ForwardSDE(Ex4(d=d, m=m)).to(device)
+    sde = ForwardSDE(problems.Ex4(d=d, m=m)).to(device)
     return sde, t, y, a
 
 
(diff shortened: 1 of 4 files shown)


tests/test_strat.py
```diff
@@ -101,18 +105,113 @@ def check_efficiency():
...
+
+        swiss_knife_gradcheck(func_high_order, inputs, modules, rtol=rtol, atol=atol, eps=eps, grad_params=True)
 
 
 test_dg_ga_jvp()
 check_efficiency()
-test_adjoint()
+
+test_adjoint_inputs()
+test_adjoint_params()
```
patrick-kidger, line 217: Would it be possible to remove these calls at the bottom and just make this operate through pytest? The tests all already start with `test_` so they should be picked up. (Not sure how you want to handle `check_efficiency`, that's really more a diagnostic than a test.)

tests/test_strat.py
```diff
@@ -101,18 +105,113 @@ def check_efficiency():
...
+                          eps: float = 1e-6,
+                          atol: float = 1e-5,
+                          rtol: float = 1e-3,
+                          grad_inputs=False,
+                          gradgrad_inputs=False,
+                          grad_params=False,
+                          gradgrad_params=False):
+    """Check grad and grad of grad wrt inputs and parameters of Modules.
+
+    Assumes func outputs a single scalar. Aimed to be as self-contained as
```
patrick-kidger, line 145: If this assumption is made, can there be an assert somewhere that validates that this is the case?

tests/test_strat.py
```diff
@@ -101,18 +105,113 @@ def check_efficiency():
...
+    swiss_knife_gradcheck(func, y0, sde, eps=1e-7, rtol=1e-3, atol=1e-3, grad_inputs=True, gradgrad_inputs=True)
+
+
+def test_adjoint_params():
+    sde = problems.Ex5(d=d, m=m, sde_type=SDE_TYPES.stratonovich).to(device)
+    bm = BrownianInterval(t0=t0, t1=t1, shape=(batch_size, m), dtype=dtype, device=device)
+
+    def func(inputs, modules):
+        """Outputs gradient norm squared."""
+        y0, sde = inputs[0], modules[0]
```
patrick-kidger, line 126: So this is because `inputs` and `modules` are expected to be 1-tuples, right? Can there be asserts validating that this is the case?

tests/test_strat.py
```diff
@@ -101,18 +105,113 @@ def check_efficiency():
...
+                plus_eps = func(inputs, modules)
+                flat_param[i] -= eps
+
+                flat_param[i] -= eps
+                minus_eps = func(inputs, modules)
+                flat_param[i] += eps
+
+                numerical_grad.append((plus_eps - minus_eps).detach() / (2 * eps))
+                del plus_eps, minus_eps
+        numerical_grad = torch.stack(numerical_grad)
```
patrick-kidger, line 196: Can this computation be wrapped in a `torch.no_grad()`?

tests/test_strat.py
lxuechen: Done.

tests/test_strat.py
lxuechen: Make it work with vector-valued functions just now. It's kinda in the style of `torch.autograd.gradgradcheck`, where a random v is used.

tests/test_strat.py
lxuechen: The input format is easy to check. However, it's hard to check that the input`func` has this type of input format. 

I noted this in the docstring.

tests/test_strat.py
lxuechen: I don't see that strictly necessary. Though, I updated where detach is placed.

lxuechen:
As a side comment, I wrote this helper function mostly since it's not straightforward to test gradients wrt `nn.Parameter` using the stuff in `torch.autograd`.

patrick-kidger:
LGTM

PR #58, opened 2020-09-18
Add adjoint support for general/scalar noise in the Ito case.

lxuechen:
So the primary reason that the scalar noise tests failed was that the corrected drift wasn't implemented. Just implemented it now and all tests are passing.  

I also wrapped the note about double backward in the docstring to be capped at 80, so that it's easier to read, say, when `help` is called in an interactive session.


patrick-kidger:
LGTM!

Issue #60, opened 2020-09-22
On BrownianInterval/Path/Tree


patrick-kidger:
@lxuechen soliciting your thoughts on something I've been contemplating, namely to make BrownianPath and BrownianTree special cases of BrownianInterval.

BrownianInterval w/ infinite cache is essentially the same as BrownianPath.
BrownianInterval w/ some tolerance at which the search is terminated, and a prespecified tree structure, is essentially the same as BrownianTree.

- Not so many Brownian* objects to support.
- Avoid the need to add Levy area support to BrownianPath/Tree, which is currently an outstanding task.
- Can remove the C++ code and just use the existing Python implementation (which seems performant enough in my experience).
- Can remove the dependence on blist, which was last updated in 2014 and has been spitting out warnings about running afoul of Python 3.9 deprecations.

Would need to add a tolerance to the search in BrownianInterval, and for robustness a trampoline to the relevant parts, neither of which should be very hard.

lxuechen:
Thanks for initiating the discussion on this! 

I agree that the Brownian portion of dev is messy. Though, I'd still like to keep BrownianPath (either C++ or Python, but ideally just one). I think it's likely I'll get it more performant on small problems than BInterval in terms of speed, since it doesn't involve the random seed splitting computation, and the Python-end of that is not too far from being finished. `blist` is slightly an issue, but I can switch to other libraries with better Python 3 support. 

I think it's also good to keep BrownianTree around, since there might be a scenario where we actually need to have a strict mapping from global entropy to the trajectory. FWIW, I don't think that's strictly achievable with BInterval without much modification, though I'm afraid I don't totally follow the proposal you listed above. If there's an easy way to let BInterval do this, then I think letting BTree be a special case sounds like a good plan. 

If the concern is that these will block the new release, I think we can temporarily deprecate the blocking ones. For now, I'm not inclined to deprecate BrownianPath. I can spend some time in getting it fixed this week, but I don't think I'll add support for davie/foster Levy area for it.

patrick-kidger:
I think it's possible to have BInterval include the global entropy->trajectory determinism when used as a BTree, so I'll make that happen.

True that BPath should be able to avoid the seed splitting overhead. I'd be inclined to have it still be an alias for BInterval, just with an additional `use_entropy` flag set to false.

[...]

Issue #54, opened 2020-09-09
Need to add OpenMP flag.


patrick-kidger:
`setup.py` doesn't currently enable OpenMP for the C++ extensions. (Which by-the-by we may wish to think about the future of as maintaining those is going to be rather a hassle.)

See e.g.
https://github.com/patrick-kidger/signatory/blob/master/setup.py#L33
for an example of how to enable the OpenMP flags.

(Whilst we're at it we should also add the fvisibility flag; see the same example.)

patrick-kidger:
Closing as we don't currently use C++.

Issue #22, opened 2020-08-16
Slight refactor, eventually


patrick-kidger:
Not suggesting that this happen now, so I'm just opening an issue to keep track of it, but I think the current hierarchy of:
```
_brownian
_core
    methods
brownian_lib
_brownian_lib [from C++]
```
(not counting `_core/adjoint_sdes` which I know you've removed in your latest branch)

should be switched to:
```
_impl
    brownian
    methods
    core
```
(So additionally tidying the Brownian things together somehow. If we end up implementing Brownian-everything in C++ then that would suffice, for example.)

lxuechen:
Agree. This refactoring seems sensible, and should come in when we have the C++ implementations.

PR #61, opened 2020-09-24
Brownian unification

patrick-kidger:
As per #60. (Also obsoletes #54, a great many things in #15, and a bit of #22)

Merged BrownianPath/Tree/Interval. BrownianPath and Tree are wrappers for BrownianInterval with particular arguments. In particular this means all of them are now on the same interface, providing the same functionality, and of course this is much easier to support. Plus, no C++ or blist cluttering things up.

This also includes several improvements to BrownianInterval:
- Use of trampolining and tail recursion
- Fewer SeedSequences are now spawned
- Those SeedSequences that are spawned now use a more efficient `spawn_key`, which should resolve the previous issue that tuple additions were taking up an appreciable fraction of the run time. The spawn keys are now a 2-tuple of `(depth in tree, index at depth level)`, rather than being long tuples that get longer each time we split.

There are some small downsides:
- We're no longer as efficient for point-based queries. The efficiency drop should be _relatively_ small: the increase in cost is logarithmic in `(length of the interval / average step size)`, and moreover this doesn't affect us in torchsde anyway as we only ever made interval-based queries.
- We're not longer running at C++ speeds. Practically speaking the difference doesn't seem to be huge: on the `sdeint_adjoint` benchmark,
    Old+Path+CPP=3.2s vs New+Path+Py=4s.
    Old+Tree+CPP=5s vs New+Tree+Py=8s.
It's also interesting to see some of the Python speedups:
    Old+Tree+Py=25s vs New+Tree+Py=8s.
    Old+Interval+Py=13s vs New+Tree+Py=7.8s
(Incidentally this is Tree w/ tol=1e-5 rather than the default 1e-6.)
If we do want to return to C++ then I'd suggest using the current BrownianInterval code as a model.

Overall I definitely think the upsides outweight the downsides.


lxuechen:
Thanks for the update. 

This PR is longer than I expected, and I'll try my best to go through it in detail soon, but it's possible that it may come after the ICLR deadline. 

A few non-ideal changes I notice already are
1. the breaking of backwards compatibility, and
2. the removal of the Brownian base class. 

1 is typically something that drives people mad, so it's essential that we support the old API. 2 is also not ideal in the long run, IMHO. A base class should help tell us and people what we expect the Brownian stuff to do, and I kinda view it as a header file that defines the whole API. Keeping it there would help other smart people potentially extend this if they wished, just like when you came up with `BrownianInterval`.

patrick-kidger:
1. Ah, I'd assumed from the removal of logqp that we were allowing backward-incompatible changes. I see now that you're planning on including this anyway, so sure, I can make things backward-compatible.
2. Fair point.

I'll make the changes.

patrick-kidger:
Okay, I've added BaseBrownian.

Regarding BrownianPath, BrownianTree: it's not feasible to have these simultaneously (a) be convenience wrappers for BrownianInterval with different default arguments, and (b) support the old API. It sounds like you prefer (b) to (a).

As such I see two possible options:
- Implement BPath and BTree by wrapping BInterval, which produces very minor backward-incompatible changes (for BPath, can no longer query outside [t0, t1]; `.to` would no longer function).
- Copy-paste all the old BPath/BTree code in from master, additionally switching out blist for a dictionary to keep functionality if not speed.

In either case adding an additional warning that BPath/BTree is deprecated and should be switched out for BInterval with appropriate arguments.

Does that fit with what you're thinking?

lxuechen:
Thanks for adding the base class back. 

I think my main point about backwards comp. is about the constructor method `__init__`. It would be nice if people could create BPath instances as before, with the same set of arguments. 

I agree to deprecate `.to`, and let the interpreter raise a NotImplError stating this particular method is deprecated when called. 

I agree with the wrapper, but don't agree with raising additional warnings that BPath and BTree is deprecated, as they shouldn't be. At the very least, BPath should be a core part of the brownian component, after I patch changes to the C++ code. This is by far a very easy way to get the speed up for small problems, and I don't see how 3.2s vs 4s is not a big regression in terms of speed. 

Happy to discuss more on points where we disagree.

patrick-kidger:
Okay, so, plan is to have BPath, BTree be wrappers for BInterval, using the old API. This will result in minor backward incompatibilities (no `.to`) but should be fine. The main advantage being that there is now only have one Brownian* object to maintain, rather than five.

- The reason for the deprecation warnings is because _not_ deprecating them implies that we're happy with the old APIs, which are (a) inconsistent with BInterval, (b) a sharply limited subset of what BInterval is capable of, and (c) involve a couple of arguments that now do nothing, e.g. `window_size` on BPath. To my mind it's better to point people towards the better thing to be using, i.e. to favour `BInterval(..., cache_size=None)` over BPath.
- The speed regression is really just because of the C++ -> Python switch, not because of the choice to implement BPath via BInterval. As such I anticipate that it's only temporary. My suggestion for C++ going forwards is to implement _just_ BrownianInterval.
*Footnote - to me, the speed improvement from 4s to 3.2s is not worth the maintenance overhead of another language, the attendant cross-operating-system problems/cross-PyTorch-versions problems, and the issue of maintaining parity between two implementations. Obviously you feel differently though; that's fine.*

lxuechen:
> Okay, so, plan is to have BPath, BTree be wrappers for BInterval, using the old API. This will result in minor backward incompatibilities (no .to) but should be fine. The main advantage being that there is now only have one Brownian* object to maintain, rather than five.

I mostly agree with this proposal. I don't think it hurts to support both how we previously constructed things and what we currently do, and we'll just state the recommendation is to follow the new API. 

> The speed regression is really just because of the C++ -> Python switch, not because of the choice to implement BPath via BInterval. As such I anticipate that it's only temporary. My suggestion for C++ going forwards is to implement just BrownianInterval.

I'm not entirely sure that's true. When queries exceed the cache, the reconstruction will take time, right? This is not to mention the various object constructions which will put quite some load on the interpreter. Though, I agree I hadn't taken the chance to fully optimize Bpath, and I agree that any speculation on performance not based on profiling doesn't mean much. My original point was to provide this as an option so that there's a chance to squeeze some performance out if that's ideal. 

So here's a proposal, we could do the wrapper logic, and I'll additionally add a `BPathv2` (open to recommendations for other names) implemented in C++. Maintaining parity between two implementations shouldn't be a problem, if the base class is well-defined. I'm happy to invest in the time to do this.

patrick-kidger:
> I mostly agree with this proposal. I don't think it hurts to support both how we previously constructed things and what we currently do, and we'll just state the recommendation is to follow the new API.

The old and new APIs are different. It's not possible to have `BrownianPath(new stuff)` and `BrownianPath(old stuff)` both work. If we want to support the new API wrt path/tree then I'd probably do it via classmethods on Interval.

> I'm not entirely sure that's true. When queries exceed the cache, the reconstruction will take time, right? This is not to mention the various object constructions which will put quite some load on the interpreter. Though, I agree I hadn't taken the chance to fully optimize Bpath, and I agree that any speculation on performance not based on profiling doesn't mean much. My original point was to provide this as an option so that there's a chance to squeeze some performance out if that's ideal.
> 
> So here's a proposal, we could do the wrapper logic, and I'll additionally add a `BPathv2` (open to recommendations for other names) implemented in C++. Maintaining parity between two implementations shouldn't be a problem, if the base class is well-defined. I'm happy to invest in the time to do this.

Setting `cache_size=None` gives an infinite cache and no reconstruction. This looks to give performance roughly equivalent to the old BPath, although I've only done rough checks of that fact. (Not too surprising; BPath, BInterval and BTree all ultimately use a tree based data structure, either explicitly or with blist.)
I essentially regard BInterval as a hybridisation/generalisation of both BPath and BTree, which is why it's possible to specialise back to each of them, and why I suggest BInterval as the thing to implement in C++.

patrick-kidger:
@lxuechen Returning to this now we all have a little time. Path and Tree now maintain the same interface (and pass the same tests) as before.

tests/test_brownian_interval.py
```diff
@@ -38,12 +37,16 @@
 REPS = 3
 MEDIUM_REPS = 25
 LARGE_REPS = 500
-ALPHA = 0.0001
+ALPHA = 0.00001
```
lxuechen, line 40: Why did this significance level become even smaller?

torchsde/_brownian/derived.py
```diff
@@ -0,0 +1,156 @@
...
+    @property
+    def shape(self):
+        return self.base_brownian.shape
+
+    @property
+    def levy_area_approximation(self):
+        return self.base_brownian.levy_area_approximation
+
+
+class BrownianPath(brownian_interval.BrownianInterval):
```
lxuechen, line 50: Is it possible to use some form of dependency injection here, as opposed to inheriting directly from BI? BP should inherit from BaseB, since the latter specifies the API. But inheriting from BI could be a rather strong dependency.

torchsde/_brownian/derived.py
```diff
@@ -0,0 +1,156 @@
...
+        # The extra tb, return_U, return_A arguments are all needed for compatibility with the current solvers.
+        assert not return_U
+        assert not return_A
+        out = super(BrownianPath, self).__call__(t, tb)
+        if tb is None:
+            out = out + self._w0
+        return out
+
+
+class BrownianTree(brownian_interval.BrownianInterval):
```
lxuechen, line 87: A similar depency argument goes here.

torchsde/_brownian/derived.py
```diff
@@ -0,0 +1,156 @@
...
+    To use:
+    >>> bm = BrownianPath(t0=0.0, w0=torch.zeros(4, 1))
+    >>> bm(0., 0.5)
+    tensor([[ 0.0733],
+            [-0.5692],
+            [ 0.1872],
+            [-0.3889]])
+    """
+
+    def __init__(self, t0: Scalar, w0: Tensor, window_size: int = 8):
```
lxuechen, line 64: Could we also try support the new API at the same time? Something like `__init__(self, t0, t1, shape..., w0=None, window_size=...)`? The legacy `w0` and `window_size` should mostly be there to not anger people who might have code depending on this codebase.

torchsde/_brownian/derived.py
```diff
@@ -0,0 +1,156 @@
...
+    To use:
+    >>> bm = BrownianTree(t0=0.0, w0=torch.zeros(4, 1))
+    >>> bm(0., 0.5)
+    tensor([[ 0.0733],
+            [-0.5692],
+            [ 0.1872],
+            [-0.3889]], device='cuda:0')
+    """
+
+    def __init__(self, t0: Scalar,
```
lxuechen, line 103: Similar argument about constructor goes here.

lxuechen (commented):
Thanks for bearing with my slowness during the last two weeks. 

In general the PR looks good. I'll add in the CPP version of BPath during the next weekend or two. 

Two additional comments I have is that:
- Could we switch the name of argument `shape` in constructor to `size`? This matches better with most torch Tensor construction functions (e.g. randn, zeros)? 
- Could we use lower case variable names for `W` and `H` in the constructor?

torchsde/_brownian/derived.py
patrick-kidger: Dependency injection, probably not. If you mean composition-over-inheritance, then yes that's possible - I was mostly just avoiding lots of redirections `def blah(self): return self._interval.blah()`, but I can change that.

torchsde/_brownian/derived.py
patrick-kidger: The problem then is matching up positional arguments. It ends up having to be something like `__init__(self, t0, w0, window_size, ...)`, with some complicated set of checks to try and figure out if they're using old or new style API - in the latter case then re-interpreting `w0` as `t1`, and so on.

Moreover there's no real need to do this - if someone wants to use the new-style API then they can just use BInterval directly.

patrick-kidger:
Absolutely no problem, we've all had a lot on our plate recently.

- If you're going for a C++ implementation then I'd suggest replicating BInterval. As per this PR, BInterval generalises both BPath and BTree, and I don't think there's any performance to be gained by special-casing either of them.

Actually, I think the best thing to implement would just be `_Interval`. That's the hot loop here. Wiring in the C++ also then becomes particularly easy - it's then just the private `BrownianInterval`<->`_Interval` interface that we need to worry about.

- shape vs size: agreed.

- lower case W, H: can do, although for example I've already been using lower case `h` to represent a time delta. Generally I'm happy to break Python casing conventions in favour of following mathematical conventions, but if you don't like it then I could change these to `increment` and `space_time_levy_area` (or similar) respectively.

torchsde/_brownian/derived.py
lxuechen: Yes, composition-over-inheritance is what I mean. The motivation is the old-school argument that in the case of inheritance, changing BI's API would affect BPath and BTree.

lxuechen:
> lower case W, H

I agree with the math over naming convention now.

patrick-kidger:
All done I think.
- Put the significance level back.
- Other changes as discussed.

README.md
```diff
@@ -7,19 +7,25 @@ Similar to [torchdiffeq](https://github.com/rtqichen/torchdiffeq), algorithms in
...
-
 ## Installation
+#### Command
+
 ```shell script
 pip install git+https://github.com/google-research/torchsde.git
 ```
 
+#### Requirements
+Python 3 and PyTorch >=1.6.0
```
lxuechen, line 18: I guess if `python_requires='~=3.6'`, we might as well state Python >= 3.6 here.

tests/test_adjoint.py
```diff
@@ -35,16 +35,14 @@
 dtype = torch.get_default_dtype()
 
 ito_methods = {'milstein': 'ito', 'srk': 'ito'}
-stratonovich_methods = {'midpoint': 'stratonovich', 'log_ode': 'stratonovich'}
```
lxuechen, line 38: Is the test flaky with log_ode?

tests/test_adjoint.py
```diff
@@ -35,16 +35,14 @@
...
 
 
 @pytest.mark.parametrize("sde_cls", [Ex1, Ex2, Ex3, Ex4])
 @pytest.mark.parametrize("method, sde_type", itertools.chain(ito_methods.items(), stratonovich_methods.items()))
 @pytest.mark.parametrize('adaptive', (False,))
 def test_adjoint(sde_cls, method, sde_type, adaptive):
     # Skipping below, since method not supported for corresponding noise types.
-    if method == METHODS.log_ode_midpoint and sde_cls.noise_type == NOISE_TYPES.diagonal:
-        return
     if method == METHODS.milstein and sde_cls.noise_type == NOISE_TYPES.general:
```
lxuechen, line 46: Optional: Not super important, but I guess we could merge these two if checks.

lxuechen:
Thanks again for the shape->size change! Looking at the diff, the task seems like a slog.

patrick-kidger:
- Tweaks as suggested.
- I removed the log-ODE test because it's quite slow: ~500 seconds, which is about 1/3 of the total runtime of the tests. On that note, I don't think we actually have tests that check that all the different solvers.
- shape->size was easy, my IDE did it for me!
- FYI I'm putting together a PR atm to update the README.

lxuechen:
> shape->size was easy, my IDE did it for me!

PR #63, opened 2020-09-24
Dev minor

lxuechen:
Minor changes
- state requirements in README,
- trim `latent_sde.py` in `examples`,
- use `torch.split` for `flat_to_shape`, since it's slightly faster than tensor slicing, and
- update documentation for `sdeint` and `sdeint_adjoint`.

This PR is mostly trigger by #61, adding some minor changes that I had in my in mind for a while. 

Even though `logqp` is removed from the documentation, I still plan to support `logqp` in the API, merely for good backward compatibility. The concrete plan there is to have a wrapper that creates a new set of `f` and `g` from existing ones in the SDE object passed in. That should come in a future PR and shouldn't be complicated as a change.


README.md
```diff
@@ -7,6 +7,14 @@ Similar to [torchdiffeq](https://github.com/rtqichen/torchdiffeq), algorithms in
...
 </p>
 
+## Requirements
+- torch>=1.6.0
+- blist
+- numpy>=1.19.1
+- boltons>=20.2.1
+
+These requirements are specified in `setup.py` using `install_requires`.
+
```
patrick-kidger, line 17: Usually I wouldn't state requirements in the readme unless some action is necessary on the user's part. (Additionally this means we have two identical lists to maintain going forward.)

examples/latent_sde.py
```diff
@@ -342,26 +390,24 @@ def main():
...
-    parser.add_argument('--show-percentiles', type=utils.str2bool, default=True)
-    parser.add_argument('--show-arrows', type=utils.str2bool, default=True)
-    parser.add_argument('--show-mean', type=utils.str2bool, default=False)
-    parser.add_argument('--hide-ticks', type=utils.str2bool, default=False)
+    parser.add_argument('--show-prior', type=str2bool, default=True)
+    parser.add_argument('--show-samples', type=str2bool, default=True)
+    parser.add_argument('--show-percentiles', type=str2bool, default=True)
+    parser.add_argument('--show-arrows', type=str2bool, default=True)
+    parser.add_argument('--show-mean', type=str2bool, default=False)
+    parser.add_argument('--hide-ticks', type=str2bool, default=False)
```
patrick-kidger, line 398: What's the rationale for not using "store_true/false"?

README.md
lxuechen: I don't see this as a big problem. I'm happy to put in the work to maintain this if you find it disturbing.

examples/latent_sde.py
lxuechen: It makes life easier when I need to have small sweeps using shell scripts, e.g. I could now just do 
```shell script
for arg in True False; do 
... 
done
```
I just updated it so that now it also works for arguments with no value passed in, e.g. both `--adjoint` and `--adjoint True` work.

README.md
patrick-kidger: More a neatness thing than anything really. Can we maybe put them lower down? Basically, they're the first thing people see when opening the README, and thus implicitly the most important part of the README, which seems wrong.

I'd suggest promoting the GBM example (of a couple sections down) to this spot.

lxuechen: I still don't see it as a problem to include these at the beginning of the README page. It shouldn't hurt to list requires in companion with installation guide.

patrick-kidger: Okey dokey then.

patrick-kidger:
Yep, LGTM.
(And welcome back, btw.)

PR #65, opened 2020-09-24
Dev adjoint strat diag

lxuechen:
The last missing pieces of adjoints:
- Stratonovich diagonal, scalar, additive
- Ito general 

I'll send in another PR later to unify test_strat and test_adjoint.

### DIFF
--- a/tests/test_adjoint.py
+++ b/tests/test_adjoint.py
@@ -80,13 +80,13 @@ def test_adjoint(problem, method, sde_type, noise_type, adaptive):
         problem.zero_grad()
         _, y1 = torchsde.sdeint(problem, y0, ts, bm=bm, method=method, adaptive=adaptive)
         y1.backward(v)
-        grad_true = torch.cat([p.grad for p in problem.parameters() if p.grad is not None])
+        grad_true = torch.cat([p.grad.reshape(-1) for p in problem.parameters() if p.grad is not None])
         del y1
 
     problem.zero_grad()
     _, y1 = torchsde.sdeint_adjoint(problem, y0, ts, bm=bm, method=method, adaptive=adaptive)
     y1.backward(v)
-    grad_adjoint = torch.cat([p.grad for p in problem.parameters() if p.grad is not None])
+    grad_adjoint = torch.cat([p.grad.reshape(-1) for p in problem.parameters() if p.grad is not None])
 
     assert_allclose(grad_true, grad_adjoint)
 
--- a/tests/test_adjoint.py
+++ b/tests/test_adjoint.py
@@ -49,20 +49,19 @@ def test_adjoint(problem, method, sde_type, noise_type, adaptive):
         return
     if problem is not Ex3 and noise_type == 'additive':
         return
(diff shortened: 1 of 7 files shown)


tests/test_adjoint.py
```diff
@@ -33,36 +33,30 @@
 device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
 dtype = torch.get_default_dtype()
 
-ito_methods = {'euler': 'ito',
-               'milstein': 'ito',
-               'srk': 'ito'}
-stratonovich_methods = {'midpoint': 'stratonovich',
-                        'log_ode': 'stratonovich'}
```
patrick-kidger, line 40: What's the rationale for removing Euler?

tests/test_adjoint.py
```diff
@@ -33,36 +33,30 @@
...
 def test_adjoint(problem, method, sde_type, noise_type, adaptive):
-    if method == 'euler' and adaptive:
-        return
     if problem is not Ex3 and noise_type == 'additive':
         return
-    # TODO: remove this once we have adjoint implemented for other noise/sde combinations
-    if sde_type == 'stratonovich' and noise_type != 'general':
-        return
     if sde_type == 'ito' and noise_type == 'general':
         return
```
patrick-kidger, line 48: Can this check also be removed now if Ito+general is supported?

tests/test_adjoint.py
```diff
@@ -33,36 +33,30 @@
...
+    if noise_type == "diagonal" and method == "log_ode":
+        return
```
patrick-kidger, line 50: So I get that log_ode is the same as midpoint when the noise is non-general. If that's the reason for this check then all non-general noise_types should trigger this.

That said, I'd prefer to test log-ode anyway. It is a valid combination.

tests/test_adjoint.py
```diff
@@ -33,36 +33,30 @@
...
 
-    d = 1 if noise_type == 'scalar' else 10
+    d = 10
+    m = {"scalar": 1}.get(noise_type, d)  # TODO: Decouple d from m.
```
patrick-kidger, line 53: can this just be `m = 1 if noise_type == 'scalar' else 10`?

torchsde/_core/base_sde.py
```diff
@@ -131,7 +131,7 @@ def dg_ga_jvp_column_sum_v1(self, t, y, a):
 
     def dg_ga_jvp_column_sum_v2(self, t, y, a):
         # Faster, but more memory intensive.
-        requires_grad = torch.is_grad_enabled()
+        requires_grad = torch.is_grad_enabled() and (t.requires_grad or y.requires_grad or a.requires_grad)
```
patrick-kidger, line 134: This probably shuldn't be the case, as per our discussion on my previous PR?

patrick-kidger (commented):
I notice that you're including a few of my old commits as well. Can those be dropped?

Overall, very cool to see that support for this is in now.

tests/test_adjoint.py
lxuechen: It's just hard for gradients to match up.

PR #66, opened 2020-10-08
Automates torch installation

Raahul-Singh:
Currently, having torchsde as a dependency creates problem while installing in isolated environments.
The problem is when installing a repo, it collects all the dependencies before it installs them. TorchSDE while being collected checks for torch. This creates problems in isolated envs as torch would not be there by default.

In this PR, I have removed the raising of the exception with a warning and have dynamically installed torch at runtime.
This should solve the build problems.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,14 +14,21 @@
 
 import os
 import platform
-
+import warnings
 import setuptools
+import subprocess
+import sys
+from importlib import reload
 
 try:
     import torch
     from torch.utils import cpp_extension
 except ModuleNotFoundError:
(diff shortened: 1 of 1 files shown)


Raahul-Singh:
> @googlebot I signed it!

@googlebot I signed it!

Raahul-Singh:
@googlebot I fixed it.

patrick-kidger:
Hi - thanks for your contribution.
As it turns out, this has already been fixed on the dev branch. If this is an issue for you now (before we make our next release) then you can install from the dev branch instead; usual stability caveats apply.

Raahul-Singh:
Awesome! Looking forward to the release! Thanks!

PR #67, opened 2020-10-13
Dev adjoint strat

lxuechen:
This is a cleaned up version of #65, and you may ignore that PR from now on. I'm only leaving that there for reference. It should be removed after this PR is merged. 

Also removed the unwanted tests and removed the analytical computation. 

The test problems module was originally designed such that each problem is of a particular noise type. I don't remember since when most problem started to support noise types that they weren't originally meant for, but I'm changing this back now. This also gets rid of the ugly and error-prone checks for Ex3 not of additive noise. 

I'm also excluding test for the case of adaptive==True. Adaptive solvers don't affect the adjoint gradients much, but affect the finite-diff gradients.


tests/test_adjoint.py
```diff
@@ -22,46 +22,46 @@
...
-@pytest.mark.parametrize("noise_type", ['diagonal', 'scalar', 'additive', 'general'])
-@pytest.mark.parametrize('adaptive', (False, True))
-def test_adjoint(problem, method, sde_type, noise_type, adaptive):
-    if method == 'euler' and adaptive:
-        return
-    if problem is not Ex3 and noise_type == 'additive':
+@pytest.mark.parametrize('adaptive', (False,))
+def test_adjoint(sde_cls, method, sde_type, adaptive):
+    # Skipping below, since method not supported for corresponding noise types.
+    if method == METHODS.log_ode_midpoint and sde_cls.noise_type == NOISE_TYPES.diagonal:
```
patrick-kidger, line 46: So I've just checked, and indeed log-ODE states that it doesn't supports diagonal noise.
I think that's a mistake - once fixed this check can also be removed.

tests/test_adjoint.py
```diff
@@ -71,28 +71,23 @@ def test_adjoint(problem, method, sde_type, noise_type, adaptive):
...
+        ys = torchsde.sdeint_adjoint(sde, y0, ts, bm, dt=dt, method=method, adaptive=adaptive)
+        return (ys[-1] ** 2).sum(dim=1).mean(dim=0)
+
+    # `grad_inputs=True` also works, but we really only care about grad wrt params and want fast tests.
+    utils.gradcheck(func, y0, sde, eps=1e-6, rtol=1e-2, atol=1e-3, grad_params=True)
 
 
 @pytest.mark.parametrize("problem", [BasicSDE1, BasicSDE2, BasicSDE3, BasicSDE4])
 @pytest.mark.parametrize("method", ito_methods.keys())
 @pytest.mark.parametrize('adaptive', (False, True))
```
patrick-kidger, line 89: Should adaptive=True be removed here? (Given that the Euler check is removed below.)

tests/problems.py
```diff
@@ -30,27 +32,11 @@
...
-    def __init__(self, d=10, sde_type=SDE_TYPES.ito, noise_type=NOISE_TYPES.diagonal):
-        super(Ex1, self).__init__(sde_type=sde_type, noise_type=noise_type)
-        self.f = self.f_ito if sde_type == SDE_TYPES.ito else self.f_stratonovich
-        self.g = {NOISE_TYPES.diagonal: self.g,
-                  NOISE_TYPES.scalar: _scalar(self.g),
-                  NOISE_TYPES.general: _general(self.g)}[noise_type]
+    noise_type = NOISE_TYPES.diagonal
+
+    def __init__(self, d, sde_type=SDE_TYPES.ito, **kwargs):
+        super(Ex1, self).__init__(sde_type=sde_type, noise_type=Ex1.noise_type)
```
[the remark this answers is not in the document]
lxuechen: I'm making `noise_type` also a class attribute so that I could decide the value of m before instantiating the object in `test_adjoint`.

tests/test_adjoint.py
lxuechen: SGTM.

tests/test_adjoint.py
lxuechen: `adaptive=True` shouldn't affect that test. I removed it for the test above since it affects the finite-difference gradients. The above tests pass if I use the analytical solutions. But not every problem has an analytical solution, and maintaining that code might become tricky.

tests/problems.py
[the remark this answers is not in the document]
patrick-kidger: Ah, I see. This is just a thing to make BaseSDE happy.

PR #68, opened 2020-10-13
Continuous Integration. 

lxuechen:


setup.py
```diff
@@ -67,7 +67,7 @@
                        optional=True)
     ],
     cmdclass={'build_ext': cpp_extension.BuildExtension},
-    install_requires=['torch>=1.6.0', 'blist', 'numpy>=1.19.1', 'boltons>=20.2.1'],
+    install_requires=['torch>=1.6.0', 'blist', 'numpy>=1.19.1', 'boltons>=20.2.1', 'scipy'],
```
patrick-kidger, line 70: can we specify an explicit version number for scipy?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,45 @@
...
+
+name: Python package
+
+on:
+  push:
+    branches: [ master, dev ]
+  pull_request:
+    branches: [ master, dev ]
+  schedule:
+  - cron: "0 2 * * 0-6"
```
patrick-kidger, line 12: Is it really necessary to test every day? Nothing actually changes most days, and that'll add up to a lot of CO2.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.6, 3.8]
+        os: [ubuntu-latest, macOS-latest]
```
patrick-kidger, line 22: why not Windows?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,45 @@
...
+    branches: [ master, dev ]
+  schedule:
+  - cron: "0 2 * * 0-6"
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      max-parallel: 4
```
patrick-kidger, line 19: Why limit the parallelism?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Test with pytest
+      run: |
+        python setup.py develop
+        pytest tests
```
patrick-kidger, line 45: I'd suggest adding a few more things to make these more robust.

When opening a PR against master, it's worth having a job that checks the version of new against old, and that new is strictly greater than old. (And condition the running of the tests on this first check passing.)

I think it's worth having another part of this step be to check that you _can't_ import torchsde before you run `setup.py`. Given that we've just cloned it down, with the right directory structure this can be possible, so this serves to verify that we're actually testing the installation at all.

If you want examples of both of these btw then have a look at [this](https://github.com/patrick-kidger/signatory/blob/master/.github/workflows/build.yml), where there's also a whole bunch of other things going on, but you should be able to pick out the bits I've just mentioned.

Incidentally I'm curious if macOS is actually compiling the C++. In my experience compiling C++ on macOS via Actions requires setting up a few extra things; see the above link for that as well. I think there should be another check that the C++ is actually compiling, so that we know that we're actually testing it.

setup.py
lxuechen: Done.

.github/workflows/python-package.yml
lxuechen: Done.

.github/workflows/python-package.yml
lxuechen: Something weird is with blist on windows. I'll add this back after merging #61.

.github/workflows/python-package.yml
lxuechen: > When opening a PR against master, it's worth having a job that checks the version of new against old, and that new is strictly greater than old. (And condition the running of the tests on this first check passing.)

I simplified your code a little and this is done. 

> I think it's worth having another part of this step be to check that you can't import torchsde before you run setup.py. Given that we've just cloned it down, with the right directory structure this can be possible, so this serves to verify that we're actually testing the installation at all.

If the goal is to make sure we're testing the installed package, then `cd ..` would suffice. 

> Incidentally I'm curious if macOS is actually compiling the C++. In my experience compiling C++ on macOS via Actions requires setting up a few extra things; see the above link for that as well. I think there should be another check that the C++ is actually compiling, so that we know that we're actually testing it.

We can come back at this after #61.

.github/workflows/python-package.yml
lxuechen: Removed.

patrick-kidger:
LGTM

PR #71, opened 2020-10-15
Added documentation

patrick-kidger:
I still need to update the examples (of the examples folder, not the examples in the README) to reflect the new world order, but otherwise I think this is all the documentation done. Opening this to receive comments on what I've done so far.

I wrote most of this from scratch so let me know if you disagree with anything / if I missed something.


README.md
```diff
@@ -1,125 +1,79 @@
-# PyTorch Implementation of Differentiable SDE Solvers ![Python package](https://github.com/google-research/torchsde/workflows/Python%20package/badge.svg?branch=dev)
-This codebase provides [stochastic differential equation (SDE)](https://en.wikipedia.org/wiki/Stochastic_differential_equation) solvers with GPU support and efficient sensitivity analysis.
-Similar to [torchdiffeq](https://github.com/rtqichen/torchdiffeq), algorithms in this repository are fully supported to run on GPUs.
+<h1 align='center'>torchsde</h1>
```
lxuechen, line 1: Could we leave the title + figure + build check badge portion of this README as before?

README.md
```diff
@@ -1,125 +1,79 @@
...
+
+sde = SDE()
+y0 = torch.full((batch_size, state_size), 0.1)
+ts = torch.linspace(0, 1, t_size)
+# Initial state y0, the SDE is solved over the interval [ts[0], ts[-1]].
+# ys will have shape (t_size, batch_size, state_size)
+ys = torchsde.sdeint(sde, y0, ts)
+```
+
+### Notebook
```
lxuechen, line 59: I think it might be worthwhile to put Notebook and LatentSDE under a section titled `examples`. You can then put your SDE-GAN model also under this section later as a subsection/paragraph, if you feel like it.

DOCUMENTATION.md
```diff
@@ -0,0 +1,170 @@
+# Documentation
```
lxuechen, line 1: This doc is pretty informative and I really appreciate it. 

No pressure in doing the following, but I think we should be able to auto-generate something like this based on just the docstring. I just tried pydoc-markdown, and it seems pretty feasible. The slight caveat there is that we would need to tweak a config file to exclude some of the private modules. This also gives us a chance to clean up the docstring. Maintaining the doc would also be easier in the ideal scenario.

README.md
patrick-kidger: Sure.

README.md
patrick-kidger: They are already under a section titled examples?

I would like to add an SDE-GAN example as well.

DOCUMENTATION.md
patrick-kidger: Thanks.

I've used Sphinx, readthedocs, before successfully, and I agree autogeneration is the way to go, long term. I've pre-emptively claimed torchsde on readthedocs -- not a priority right now but something to aim for on the next release, I think?

lxuechen: SGTM. You could also add me to the readthedocs project, the username is the same as my github handle.

README.md
lxuechen: Ah sorry, you're right.

lxuechen:
The BI tests are a little flaky. I made some tweaks in #69, but they're perhaps not sufficient to get things working really stably.

DOCUMENTATION.md
patrick-kidger: Will do!

patrick-kidger:
Agreed. I'll dig into that. (In a separate PR).

patrick-kidger:
@lxuechen Since you beat me to updating the notebook example then I think this PR is probably good to go. I know there's a couple of outstanding to-dos that I need to sort out (BInterval tests; SDE-GAN example), but once that done I think we're ready for the next release?

lxuechen:
SGTM! I think we're almost there, but definitely feel free to let me know if there's anything else on your mind that'd be appropriate for v0.2.0.

patrick-kidger:
I'll wait to get #76 in and re-run with that, then look at merging this.

PR #72, opened 2020-10-15
Add fail-fast: false

patrick-kidger:
Set the tests not to cancel if one of them fails.

Unrelatedly, do you want to make passing the tests a requirement to merge?
Settings->Branches->Branch protection rule.
Can also use this to require reviews before merging.
Leaving "include administrators" unticked will allow (just) you to merge even if the tests fail by clicking through a "I know what I'm doing" warning.


lxuechen:
I think we already have a branch protection rule on master that requires 1 review. I can add in other rules once things are merged to master. Thanks for this reminder.

patrick-kidger:
Ah, I actually meant for the dev branch. (Less important there of course.)

PR #73, opened 2020-10-15
Dev methods fixes

patrick-kidger:
Ticking off a variety of smaller to-dos from #15 :
- Fixed adaptivity checks to not trigger for Heun etc.
- Improved default method selection to not use SRK when that's not appropriate. (Strat etc.)
- Updated `test_sdeint.py` to use pytest, and improved its test coverage.

The tests in `test_sdeint.py` are very numerous (~2000) covering all the common combinations of inputs, but are quick to run (~3 minutes); I felt it was best to be comprehensive on this central part of the library.

Thanks for your email btw - fortunately we've been doing different things, but I should have let you know!

### DIFF
--- a/tests/basic_sde.py
+++ b/tests/basic_sde.py
@@ -158,33 +158,6 @@ class ScalarSDE(AdditiveSDE):
         super(ScalarSDE, self).__init__(d=d, m=1, sde_type=sde_type)
 
 
-class TupleSDE(SDEIto):
-    def __init__(self, d=10):
-        super(TupleSDE, self).__init__(noise_type="diagonal")
-        self.shared_param = nn.Parameter(torch.randn(1, d), requires_grad=True)
-        self.no_grad_param = nn.Parameter(torch.randn(1, d), requires_grad=False)
-        self.unused_param1 = nn.Parameter(torch.randn(1, d), requires_grad=False)
-        self.unused_param2 = nn.Parameter(torch.randn(1, d), requires_grad=True)
-
-    def f(self, t, y):
-        y, = y
-        return (
-            self.shared_param * torch.sin(y) * 0.2 +
-            torch.sin(y ** 2.) * 0.1 +
-            torch.cos(t) +
-            self.no_grad_param * y,
-        )
-
-    def g(self, t, y):
-        y, = y
-        return torch.sigmoid(
-            self.shared_param * torch.cos(y) * .3 + torch.sin(t)) + torch.sigmoid(self.no_grad_param * y),
-
-    def h(self, t, y):
-        y, = y
-        return torch.sigmoid(y),
-
-
 class CustomNamesSDE(SDEIto):
     def __init__(self):
         super(CustomNamesSDE, self).__init__(noise_type="diagonal")
--- a/tests/basic_sde.py
+++ b/tests/basic_sde.py
@@ -94,9 +94,12 @@ class BasicSDE4(SDEIto):
         return torch.sigmoid(y)
 
 
-class GeneralSDE(SDEIto):
-    def __init__(self, d=10, m=3):
-        super(GeneralSDE, self).__init__(noise_type="general")
+class GeneralSDE(torch.nn.Module):
+    noise_type = 'general'
+
+    def __init__(self, d=10, m=3, sde_type=SDE_TYPES.ito):
+        super(GeneralSDE, self).__init__()
+        self.sde_type = sde_type
         self.shared_param = nn.Parameter(torch.randn(1, d), requires_grad=True)
         self.no_grad_param = nn.Parameter(torch.randn(1, d, m), requires_grad=False)
         self.unused_param1 = nn.Parameter(torch.randn(1, d), requires_grad=False)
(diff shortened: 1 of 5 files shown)


tests/test_sdeint.py
```diff
@@ -35,152 +36,149 @@
...
+    else:
+        if method in ('euler', 'srk'):
+            should_fail = True
+    if method in ('milstein', 'srk') and sde_cls.noise_type == 'general':
+        should_fail = True
+    if method == 'srk' and levy_area_approximation == 'none':
+        should_fail = True
+    if method == 'log_ode' and levy_area_approximation in ('none', 'space-time'):
+        should_fail = True
+    if method == 'blah':
```
lxuechen, line 109: I think we could save ourselves from checking `method=='blah'` if we did the not-in-method-list-checks above.

tests/test_sdeint.py
```diff
@@ -35,152 +36,149 @@
...
+        ctx = pytest.warns(UserWarning)
+    else:
+        ctx = _nullcontext()
+
+    # Using `f` as drift.
+    with torch.no_grad():
+        try:
+            with ctx:
+                ans = torchsde.sdeint(sde, y0, ts, bm, method=method, dt=dt, adaptive=adaptive)
+        except Exception:
```
lxuechen, line 165: There's something subtle here which is the exception raised might not be the exception we expect. I think catching an exception with a more specific type will help. Most cases that we raise explicitly are of ValueError type.

tests/test_sdeint.py
```diff
@@ -35,152 +36,149 @@
...
+    bm = None
+    logqp = False
+    should_fail = False
+    _test_sdeint(sde, bm, method, adaptive, logqp, device, should_fail)
+
+
+def _test_sdeint(sde, bm, method, adaptive, logqp, device, should_fail):
+    y0 = torch.ones(batch_size, d, device=device)
+    ts = torch.linspace(t0, t1, steps=T, device=device)
+    if adaptive and method == 'euler' and sde.noise_type != 'additive':
```
lxuechen, line 155: Doing this userwarning check here is probably going a bit far. This if-check is basically the exact same code as in `sdeint.py`, so I'd think it's unlikely something will go wrong. I wouldn't recommend including it just to keep things simple and easily maintainable. Not a super strong feeling though.

lxuechen:
I think we should also be able to merge `problems.py` and `basic_sde.py` now. I'd suggest keeping Ex1-4 and BasicSDE1-4. No pressure in doing it in this PR. I can also send in another PR on this issue.

tests/basic_sde.py
```diff
@@ -94,9 +94,12 @@ def h(self, t, y):
         return torch.sigmoid(y)
 
 
-class GeneralSDE(SDEIto):
-    def __init__(self, d=10, m=3):
-        super(GeneralSDE, self).__init__(noise_type="general")
+class GeneralSDE(torch.nn.Module):
```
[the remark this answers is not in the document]
patrick-kidger: No, because that requires using `noise_type` as an instance attribute - here it's neater to have it as a class attribute.

tests/basic_sde.py
```diff
@@ -129,10 +152,10 @@ def h(self, t, y):
...
-        super(ScalarSDE, self).__init__(d=d, m=m)
-        self.g_param = nn.Parameter(torch.sigmoid(torch.randn(1, d, 1)), requires_grad=True)
-        self.noise_type = "scalar"
+    noise_type = "scalar"
+
+    def __init__(self, d=10, sde_type=SDE_TYPES.ito):
+        super(ScalarSDE, self).__init__(d=d, m=1, sde_type=sde_type)
 
 
 class TupleSDE(SDEIto):
```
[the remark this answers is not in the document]
patrick-kidger: Sure.

torchsde/_core/sdeint.py
```diff
@@ -208,7 +222,7 @@ def integrate(sde, y0, ts, bm, method, dt, adaptive, rtol, atol, dt_min, options
         dt_min=dt_min,
         options=options
     )
-    if adaptive and solver.strong_order < 1.0:
-        warnings.warn(f"Numerical solution is only guaranteed to converge to the correct solution "
-                      f"when a strong order >=1.0 scheme is used for adaptive time-stepping.")
+    if adaptive and method == METHODS.euler and sde.noise_type != NOISE_TYPES.additive:
+        warnings.warn(f"Numerical solution is not guaranteed to converge to the correct solution when using adaptive "
+                      f"time-stepping with the Euler--Maruyama method with non-additive noise.")
```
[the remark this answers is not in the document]
patrick-kidger: Double hyphen corresponds to an en dash in latex, so I use it as a reasonable approximation to an en dash when writing code.

Heun's method should converge, yes. AFAIK it's really only Euler's method that things go wrong for.

tests/test_sdeint.py
```diff
@@ -35,152 +36,149 @@
...
+    _test_sdeint(sde, bm, method, adaptive, logqp, device, should_fail)
+
+
+@pytest.mark.parametrize("sde_cls", [basic_sde.BasicSDE1, basic_sde.BasicSDE2, basic_sde.BasicSDE3,
+                                     basic_sde.BasicSDE4])
+@pytest.mark.parametrize('method', ['euler', 'milstein', 'srk'])
+@pytest.mark.parametrize('adaptive', [False, True])
+@pytest.mark.parametrize('device', devices)
+def test_sdeint_dependencies(sde_cls, method, adaptive, device):
+    # This test uses diagonal noise. This checks if the solvers still work when some of the functions don't depend on
```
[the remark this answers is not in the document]
patrick-kidger: Done.

tests/test_sdeint.py
```diff
@@ -35,152 +36,149 @@
...
+    # (a) runs/raises an error as appropriate
+    # (b) produces tensors of the right shape
+    # (c) accepts every method
+
+    if logqp:
+        pytest.skip("Temporarily deprecating logqp.")
+
+    should_fail = False
+    if sde_type == 'ito':
+        if method in ('euler_heun', 'heun', 'midpoint', 'log_ode'):
```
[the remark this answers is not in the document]
patrick-kidger: Done.

tests/test_sdeint.py
patrick-kidger: Done.

tests/test_sdeint.py
patrick-kidger: Yup, you're right. I've now also tweaked a couple asserts to now raise ValueErrors instead.
(Shouldn't be a backward compatibility issue, I think the asserts aren't in master.)

tests/test_sdeint.py
patrick-kidger: This is mostly just so that the warning doesn't bubble up to get reported at the pytest level.
(I originally had a `warnings.simplefilter` on this, but figured if I was going to include a context like that then I might as well do it properly.)

patrick-kidger:
I agree merging them makes sense. I've left them alone for now though.

PR #75, opened 2020-10-18
Dev logqp

lxuechen:
logqp backwards compatibility.


torchsde/_core/sdeint.py
```diff
@@ -92,22 +96,29 @@ def sdeint(sde: base_sde.BaseSDE,
...
     missing_funcs = [func for func in required_funcs if not hasattr(sde, func)]
     if len(missing_funcs) > 0:
         raise ValueError(f"sde is required to have the methods {required_funcs}. Missing functions: {missing_funcs}")
 
+    # --- Backwards compatibility: v0.1.1. ---
+    if logqp:
+        sde = base_sde.SDELogqp(sde)
+        y0 = torch.cat((y0, y0.new_zeros(size=(y0.size(0), 1))), dim=1)
+    # ----------------------------------------
+
```
patrick-kidger, line 121: This new block should be moved below the noise type and sde type checks, as `SDELogqp` relies on noise type existing.

patrick-kidger (commented):
It would be nice to add a test that verifies that logqp is computing the thing it claims to be.
Other than that, this looks pretty good.

torchsde/_core/sdeint.py
lxuechen: Good point! Done.

lxuechen:
> It would be nice to add a test that verifies that logqp is computing the thing it claims to be.
> Other than that, this looks pretty good.

This would be a little difficult to check. Given that the augmentation is quite simple, I think that the chances of there being a bug is quite low. 

I'll run the example with this logqp argument to check if the model still trains, but I don't think it's natural to include this in a test suite.

PR #81, opened 2020-10-27
Improve usability.

lxuechen:
Minor fixes to enhance usability.

### DIFF
--- a/torchsde/_brownian/derived.py
+++ b/torchsde/_brownian/derived.py
@@ -198,20 +198,9 @@ def binterval_like(y: Tensor,
                    size: Optional[Tuple[int, ...]] = None,
                    dtype: Optional[torch.dtype] = None,
                    device: Optional[Union[str, torch.device]] = None,
-                   entropy: Optional[int] = None,
-                   dt: Optional[Scalar] = None,
-                   tol: Scalar = 0.,
-                   pool_size: int = 8,
-                   cache_size: Optional[int] = 45,
-                   halfway_tree: bool = False,
-                   levy_area_approximation: str = LEVY_AREA_APPROXIMATIONS.none,
-                   W: Optional[Tensor] = None,
-                   H: Optional[Tensor] = None):
+                   **kwargs):
     """Returns a BrownianInterval object with the same size, device, and dtype as a given tensor."""
-    size = y.size() if size is None else size
+    size = y.shape if size is None else size
     dtype = y.dtype if dtype is None else dtype
     device = y.device if device is None else device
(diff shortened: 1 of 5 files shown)


torchsde/_brownian/derived.py
```diff
@@ -187,3 +190,28 @@ def shape(self):
...
+                   W: Optional[Tensor] = None,
+                   H: Optional[Tensor] = None):
+    """Returns a BrownianInterval object with the same size, device, and dtype as a given tensor."""
+    size = y.size() if size is None else size
+    dtype = y.dtype if dtype is None else dtype
+    device = y.device if device is None else device
+    return brownian_interval.BrownianInterval(
+        t0=t0, t1=t1, size=size, dtype=dtype, device=device, entropy=entropy, dt=dt, tol=tol, pool_size=pool_size,
+        cache_size=cache_size, halfway_tree=halfway_tree, levy_area_approximation=levy_area_approximation, W=W, H=H
+    )
```
patrick-kidger, line 217: As this is a class constructor I'd prefer to put it as a class method on the class.
Moreover it'd be better not to repeat all the default arguments, as there's now no longer a single source of truth.

Instead:
```python
class BrownianInterval:
    ...
    @classmethod
    def like(cls, y, **kwargs):
        try:
            size = kwargs.pop('size')
        except KeyError:
            size = y.size()
        ...dtype and device...
        return cls(size=size, dtype=dtype, device=device, **kwargs)
```

patrick-kidger (commented):
Btw, do you want to open this against dev? I'll put the example there too, as I expect we'll want to combine release version.

torchsde/_brownian/derived.py
lxuechen: I think it's better to make it just a function by itself. `torchsde.binterval_like` is much faster to type and read than `torchsde.BrownianInterval.like`. 

Much of this is inspired by the frequent use of `torch.randn_like`.

I modified the arguments so that only the most commonly used are type annotated, and the others are collapsed into `**kwargs`.

patrick-kidger: Hmm, I don't agree, but okay. I would then ask that it be `brownian_interval_like`, though.

This function is now introducing defaults for `t0` and `t1`. I can see that they're necessary in this context, but these should be added as default arguments to `BrownianInterval` as well, for consistency.

lxuechen: Sure, though I'd say I prefer to keep things simple for this one mostly as it's a short-cut that helps type fewer characters.

patrick-kidger:
LGTM

PR #84, opened 2020-12-23
Major bugfix + new efficiency-focused functionality.

patrick-kidger:
Heyhey. I've been tweaking some of the SDE-GAN stuff (for the paper, not the example), and realised that it could be made much more efficient with library support... and what's the point of being a developer of the library if not for that?

- Major bugfix: fixed calling `solver.integrate` twice in `sdeint.py::integrate` (!) (I can make this a separate PR if you want to put this in immediately.)
- Added `f_and_g`, `g_prod`, and `f_and_g_prod` as functions that the SDE can define, which will be used in the solvers if possible. `f` and `g` no longer actually have to be defined if the solver doesn't need them. [1]
- Improved efficiency:
    - logqp=True now takes advantage of this to define an `f_and_g` function that is more efficient than calling them separately. (Unrelated to what I'm doing, but why not.) [2]
    - Milstein and derivative-free Milstein now no longer have extra `g` computations.
- Improved the error checking in check_contract to explicitly require a single batch dimension.

(N.B. This PR depends upon #85.)

[1] My use case for this: First of all, I'd like drift and diffusion to just be different final layers at the end of the same MLP, hence `f_and_g`. Second, for the discriminator, the diffusion ends up being a product of matrices. Doing (matrix-matrix)-vector would be inefficient, hence `g_prod`/`f_and_g_prod`.

[2] There's still some improvements that could be made here: for simplicity logqp still requires `f`/`g`/`h` to all be specified, without exploiting `f_and_g` or `f_and_g_prod` if they're available. The logical extension of this would be to also add `f_and_g_and_h`, `f_and_g_prod_and_h` as possible methods.

### DIFF
--- a/torchsde/_core/sdeint.py
+++ b/torchsde/_core/sdeint.py
@@ -104,14 +104,10 @@ def check_contract(sde, y0, ts, bm, method, names, logqp):
     if names is None:
         names_to_change = {}
     else:
-        names_to_change = {key: names[key] for key in ("drift", "diffusion", "prior_drift") if key in names}
+        names_to_change = {key: names[key] for key in ("drift", "diffusion", "prior_drift", "drift_and_diffusion",
+                                                       "drift_and_diffusion_prod") if key in names}
     if len(names_to_change) > 0:
         sde = base_sde.RenameMethodsSDE(sde, **names_to_change)
-
-    required_funcs = ("f", "g", "h") if logqp else ("f", "g")
-    missing_funcs = [func for func in required_funcs if not hasattr(sde, func)]
-    if len(missing_funcs) > 0:
-        raise ValueError(f"sde is required to have the methods {required_funcs}. Missing functions: {missing_funcs}")
 
     if not hasattr(sde, "noise_type"):
         raise ValueError(f"sde does not have the attribute noise_type.")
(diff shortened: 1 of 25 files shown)


lxuechen:
Thanks for this update! `solver.integrate` twice is a serious bug that might have been introduced by merging commits. Could you split this into a separate PR? We could merge this to master immediately. 

Could you also split bullets 2-4 into 2 separate PRs? One for 2, and another for 3-4. That'll help a lot in terms of going through the code. 

It might be too pesky to split this into 4, but I think it's reasonable to factor out the f_and_g functionality, as I think it would affect a few things.

torchsde/_core/sdeint.py
```diff
@@ -145,63 +146,103 @@ def check_contract(sde, y0, ts, bm, method, names, logqp):
...
+            raise ValueError("State sizes not consistent.")
+    for noise_size in noise_sizes[1:]:
+        if noise_size != noise_sizes[0]:
+            raise ValueError("Noise sizes not consistent.")
+
+    if sde.noise_type == NOISE_TYPES.scalar:
+        if noise_sizes[0] != 1:
+            raise ValueError(f"Scalar noise must have only one channel; the diffusion has {noise_sizes[0]} noise "
+                             f"channels.")
+
```
patrick-kidger, line 245: This red+green block above here is the entirety of bullet point 4, on better contract checking. It's actually in this PR because it interacts with the `f_and_g` functionality, as we need to check that what that returns is sane, so I don't think it makes sense to split this into a separate PR.

I'd note that about half of the changes apparently in this PR are actually #85, which makes this PR look bigger than it is. Once that's in it'll be easier to read.

lxuechen:
For efficiency purposes, I'm committing minor tweaks so we can merge. For the record, I'm rebasing from `dev` directly after merging #85, and I added the license for `diagnostics/run_all.py`

PR #85, opened 2020-12-24
Dev diagnostics fixes

patrick-kidger:
Several improvements to the diagnostics.
- They actually run now, for one thing...
- Switched all the diagnostics from Ex* to neural networks. It turns out that the previous examples exhibited structure that meant that the reported orders weren't accurate for general problems of that type. Neural networks make convenient unstructured vector fields.
- Added a `diagnostics/run_all.py`
- Added some combinations that were missing (e.g. Euler on general noise Ito)
- Various misc. tidy up.
- Cheapened the `tests/test_sdeint.py` tests (which kind of tie into this, but can be a separate PR if you prefer) so that they run in about 1/3 the time.


diagnostics/inspection.py
```diff
@@ -24,46 +24,48 @@
 from . import utils
 
 
+sys.setrecursionlimit(5000)
```
patrick-kidger, line 27: And to head off the obvious question: I was getting some recursion errors in the "true" (dt=2^-14) simulation. That's a ludicrously small step size so I don't think it's an issue generally speaking.

tests/problems.py
```diff
@@ -133,44 +132,127 @@ def nfe(self):
...
+        return torch.zeros_like(y)
+
+
+class Neural3(BaseSDE):
+    noise_type = NOISE_TYPES.additive
 
     def __init__(self, d, m, sde_type=SDE_TYPES.ito, **kwargs):
-        super(Ex4, self).__init__(sde_type=sde_type, noise_type=Ex4.noise_type)
-        self._nfe = 0
+        super(Neural3, self).__init__(sde_type=sde_type, noise_type=Neural3.noise_type)
```
patrick-kidger, line 199: Git is choosing to interleave Ex4 and Neural3 here; Ex4 and Neural4 would have been a better match. (Just to help you understand why there seem to be lots of changes.)

diagnostics/inspection.py
lxuechen: Sounds good. A `dt` that small is almost never used in practice, though as a diagnostic for the rate, it should still be worthwhile to go this far.

diagnostics/inspection.py
```diff
@@ -24,46 +24,48 @@
...
+
 def inspect_samples(y0: Tensor,
                     ts: Vector,
                     dt: Scalar,
                     sde: BaseSDE,
                     bm: BaseBrownian,
                     img_dir: str,
                     methods: Tuple[str, ...],
-                    options: Optional[Tuple] = None,
+                    options: Tuple,
```
lxuechen, line 37: Is there a particular reason to remove the default `None` for `options` and `labels`?

diagnostics/inspection.py
```diff
@@ -24,46 +24,48 @@
...
     if labels is None:
         labels = methods
 
-    sde = copy.deepcopy(sde).requires_grad_(False)
-
-    solns = [
-        sdeint(sde, y0, ts, bm, method=method, dt=dt, options=options_)
-        for method, options_ in zip(methods, options)
-    ]
+    with torch.no_grad():
```
lxuechen, line 46: Since this whole function doesn't require grads to be computed, why not just decorate the whole func? E.g. 
```
@torch.no_grad()
def inspect_samples...
```

tests/problems.py
```diff
@@ -133,44 +132,127 @@ def nfe(self):
         return self._nfe
 
 
-class Ex4(BaseSDE):
-    noise_type = NOISE_TYPES.general
+class Neural1(BaseSDE):
```
lxuechen, line 135: Neural1,2,3 doesn't seem be the best names we could give to these classes. Could we use something that's more interpretable w/o reading the impl?

patrick-kidger:
Addressed all comments!

lxuechen:
Rebasing from master, simplified a launcher script and fixed a docstring, merging now.

PR #87, opened 2021-01-05
Fix adaptive error estimate.

lxuechen:
Fix bug caused by tuple of tensor to tensor conversion.


lxuechen:
A short but important change that fixes a bug in the adaptive solver. cc'ing @patrick-kidger to keep people informed.

Issue #96, opened 2021-05-14
Upload to PyPI


patrick-kidger:
It'd be nice to make torchsde available on PyPI. There's a few limitations with hosting it only on GitHub, most notably that (a) you can't specify a range of acceptable versions (`torchsde>=0.2.4`), and (b) it's not possible to upload any _other_ package to PyPI that depends on torchsde: PyPI disallows uploading packages that have dependencies outside of PyPI.

lxuechen:
Good call! Let's do this after neurips.

dhirschfeld:
FYI: `torchsde` is available via `conda-forgge`:
```
conda install -c conda-forge torchsde
```
https://github.com/conda-forge/torchsde-feedstock

SauravMaheshkar:
@patrick-kidger Request for this issue to be closed as the package already is available on PyPI, for reference visit [**this page**](https://pypi.org/project/torchsde/).

patrick-kidger:
Ah yep!

PR #100, opened 2021-07-20
Extend CI workflow to include PyPI dist publishing

Zymrael:
Minor PR, title says it all.

Publishing is triggered only by release events. To work, the addition requires setup of a PyPI password in repo secrets.

Addresses [this](https://github.com/google-research/torchsde/issues/96).

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -92,3 +92,14 @@ jobs:
           python setup.py develop
           cd ..
           pytest torchsde/tests
+
+      - name: Build
+        run: >-
+          python setup.py sdist bdist_wheel
+
+      - name: Publish to PyPI
+        if: github.event_name == 'release'
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_password }}
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -103,3 +103,4 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.pypi_password }}
+


Zymrael:
@googlebot I signed it!

patrick-kidger:
So I note that this will trigger on quite a lot of things, including a cron schedule. It would be preferred to have it trigger only when we merge a pull request into master; moreover it'd need to check that the version number has been incremented, to prevent PyPI failures.

That said, I'd be keen to have this happen, as it's realistic that waiting for me or @lxuechen to push a button will be a bottleneck in the process... :D

patrick-kidger:
Ah yes, you're right on both counts.

Only @lxuechen can actually add the necessary secret (the PyPI key) for this, though. @lxuechen I'm happy with this PR as written if you are; we just need to bear in mind that anything we don't want immediately published on PyPI should be merged to dev rather than master.

lxuechen:
> Ah yes, I see the if statement now. I think the if statement will never trigger though, because the `on:` section above doesn't include release events?
> 
> So I think creating a separate workflow that does this, and triggers on release events, would probably be smoothest?
> 
> Pinging @lxuechen as the only one who can actually add the necessary secret (the PyPI key) for this.

I agree with Patrick here. What's more, it seems this publishing step will be ran more than once due to the multiple run environments constructed by the build matrix. Splitting the publishing step to a separate workflow is a good idea. Enforcing dependencies among workflows is a little tricky, but can be done with [workflow_run](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run). 

I have created the necessary secret based on an API token, but might need Patrick to add me to the project on PyPI.

patrick-kidger:
Right, I think I've done everything that should be necessary to get this working. I've opened three small PRs (#102,#103,#104) to add all the necessary functionality, in place of this one.

Zymrael:
I will be closing this PR then.

PR #104, opened 2021-07-22
added publishing workflow

patrick-kidger:
Prompted by (and I expect superseding) #100, I've put together a publish workflow. It should automatically upload a build to PyPI whenever we do a release.

Untested for obvious reasons; might need tweaking for all I know.

At the moment it uses a PyPI username + password for authentication. We can change that to using an API token if you prefer.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,99 @@
+name: Publish
+on:
+  release:
+    types: [published]
+    branches: [master]
+
+jobs:
+  build_and_test:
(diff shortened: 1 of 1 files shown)


patrick-kidger:
(We should create a v0.2.5 release before merging this.)

lxuechen (approved):

lxuechen:
> (We should create a v0.2.5 release before merging this.)

We could tag the commit that comes from merging #102. Thanks again for putting all of this together!

PR #137, opened 2023-08-29
bump version number

vladmandic:
bumps version number of torchsde package so it can be used to create a new release and published to pypi.
see #131 for details

PR #140, opened 2023-09-26
Fix CI

akx:
Re the discussion on #137 (hey @vladmandic 👋), this PR attempts to modernize and fix the CI process to actually pass again.

Along the way, it:

* drops support for EOL versions of Python (because Torch isn't being shipped for EOL versions either)
* drops the Boltons dependency, which was used for a single LRU dictionary, in favor of about 10 lines of code to the same effect.

You can see this passing over at https://github.com/akx/torchsde/pull/1 😊

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -54,7 +54,7 @@ jobs:
     needs: [ check_version ]
     strategy:
       matrix:
-        python-version: [ 3.6, 3.8 ]
+        python-version: [ 3.8 ]
         os: [ ubuntu-latest, macOS-latest, windows-latest ]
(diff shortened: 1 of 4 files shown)


akx:
@patrick-kidger If you could approve running the workflows, that'd be grand 😄

vladmandic:
btw, i think git workflows also need to be updated so it doesn't try running with python 3.6 anymore.

akx:
@patrick-kidger CI's showin' all green!

patrick-kidger (approved):

patrick-kidger:
Hurrah! This LGTM. Thank you for fixing this up. Time for a new release.

PR #147, opened 2024-05-25
Kornia library compatibulity

matteoguarrera:
kornia library has refactored some functions and the example/cont_ddpm.py was not working anymore due to an error raised 
 in example/unet.py  that needs to be changed to `return kornia.filters.filter2d(x, self.f, normalized=True)`.


matteoguarrera:
minor bug in the visualization function of `latent_sde_lorentz.py`

PR #156, opened 2024-12-05
Update brownian_interval.py to allow torchsde to use torch_xla

chaowenguo1:
Update brownian_interval.py to allow torchsde to use torch_xla, torch.Generator('xla') is not working


chaowenguo1:
@lxuechen @patrick-kidger @mtsokol could you review and prove the workflow

chaowenguo:
@lxuechen @patrick-kidger @mtsokol could you review and prove the workflow

